PR #96, opened 2025-08-01
(extension) rework title mapping workflow

Mr-Quin:


Mr-Quin:
Task linked: [DA-115 Change how search mapping works](https://app.clickup.com/t/90131020449/DA-115)

PR #102, opened 2025-08-03
(extension) stop emitting danmaku when page is not visible

Mr-Quin:


Mr-Quin:
Task linked: [DA-119 Prevent emitting danmaku when tab is in background](https://app.clickup.com/t/90131020449/DA-119)

PR #103, opened 2025-08-04
(extension) only ask to save mapping if the clicked season is different [DA-118]

Mr-Quin:
…nt from the currently mapped one


Mr-Quin:
Task linked: [DA-118 Don't ask for season mapping if mapping to the same season](https://app.clickup.com/t/90131020449/DA-118)

PR #112, opened 2025-08-16
(extension) layer controller should appear on top of danmaku layer [DA-133]

Mr-Quin:


Mr-Quin:
Task linked: [DA-133 Fix danmaku layer, controller should appear on top of danmaku](https://app.clickup.com/t/90131020449/DA-133)

PR #113, opened 2025-08-16
(extension) set host styles using css instead of inline

Mr-Quin:


Mr-Quin:
Task linked: [DA-136 Use css file to set host style](https://app.clickup.com/t/90131020449/DA-136)

PR #147, opened 2025-08-28
(extension) fix crash when importing a large number of files [DA-173]

Mr-Quin:


Mr-Quin:
Task linked: [DA-173 Fix crash when importing large number of files](https://app.clickup.com/t/90131020449/DA-173)

PR #156, opened 2025-08-31
(extension) autofill name when adding mount config [DA-191]

Mr-Quin:


Mr-Quin:
Task linked: [DA-191 Autofill name when adding mount config](https://app.clickup.com/t/90131020449/DA-191)

PR #158, opened 2025-08-31
(proxy) fix deployment command [DA-194]

Mr-Quin:


Mr-Quin:
Task linked: [DA-194 Fix deployment command](https://app.clickup.com/t/90131020449/DA-194)

PR #159, opened 2025-08-31
(extension) make generic danmaku endpoint input a dropdown instead of a text field [DA-193]

Mr-Quin:


Mr-Quin:
Task linked: [DA-193 Make generic danmaku endpoint input a dropdown instead of a text field](https://app.clickup.com/t/90131020449/DA-193)

PR #172, opened 2025-09-02
Migrate danmaku to compressed storage

Mr-Quin:
Extract and compress danmaku comments into a new `danmakuData` table, referencing them from `episode` and `customEpisode` to optimize storage and ensure atomic operations.

This change introduces a new IndexedDB table (`danmakuData`) to store `comments` as gzip-compressed blobs, reducing the footprint of `episode` and `customEpisode` records. A new DB migration (v12) handles existing data. `DanmakuService` and `SeasonService` are updated to read/write/delete comments via this new table, ensuring all related operations remain atomic.

---
<a href="https://cursor.com/background-agent?bcId=bc-ab8e4db6-a370-4a32-9498-51dd7fd6ad4c">
  <picture>
    <source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/open-in-cursor-dark.svg">
    <source media="(prefers-color-scheme: light)" srcset="https://cursor.com/open-in-cursor-light.svg">
    Open in Cursor
  </picture>
</a>
<a href="https://cursor.com/agents?id=bc-ab8e4db6-a370-4a32-9498-51dd7fd6ad4c">
  <picture>
    <source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/open-in-web-dark.svg">
    <source media="(prefers-color-scheme: light)" srcset="https://cursor.com/open-in-web-light.svg">
    Open in Web
  </picture>
</a>

### DIFF
--- a/packages/danmaku-anywhere/src/background/services/DanmakuService.ts
+++ b/packages/danmaku-anywhere/src/background/services/DanmakuService.ts
@@ -13,6 +13,7 @@ import {
   zCombinedDanmaku,
 } from '@danmaku-anywhere/danmaku-converter'
 import type { SeasonService } from '@/background/services/SeasonService'
+import { Dexie } from 'dexie'
 import type {
   CustomEpisodeQueryFilter,
   DanmakuImportData,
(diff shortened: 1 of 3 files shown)


Mr-Quin:
### Sweep verdict: OBSOLETE-RECOMMEND-CLOSE (recommend closing this PR, keeping DA-182 open)

Swept as part of a full open-PR sweep against master `b457aa69`. Nothing was pushed to this branch.

**The idea is still wanted.** Master still stores comments inline and uncompressed: `common/db/db.ts:25-26` declares `episode!: Dexie.Table<Episode, ...>` / `customEpisode!: Dexie.Table<CustomEpisode, ...>`, and there is no `danmakuData` table or any `CompressionStream` use anywhere under `packages/danmaku-anywhere/src`. The `filterLite` / `filterCustomLite` added since then do not help the storage problem: both call the full `filter` first and then strip `comments` in memory, so every comment is still read out of IndexedDB. DA-182 is still `backlog` / high priority.

**But this branch cannot be rebased.** Merge-base is `d0e677ff` (#171), 297 merged PRs back. `git merge-tree` conflicts in all three files, and one is a modify/delete:

- `background/services/DanmakuService.ts` and `background/services/SeasonService.ts` were deleted by `57728df7 (extension) refactor danmaku providers into more maintainable patterns [DA-236] (#195)` and moved under `background/services/persistence/`. `SeasonService` is a full rewrite (112 lines gone at the old path, 178 new plus a test at the new one); `DanmakuService` is 239 changed lines away from the version this PR rewrites.
- The rewrite is structural. Master's `persistence/DanmakuService.ts:31-45` is `@injectable('Singleton')` with `@inject(DanmakuAnywhereDb) private db` and reaches storage as `this.db.episode`; every method this PR rewrites uses the module-level `db` singleton.
- `.version(12)` is taken: on master that number is the season/episode restructure (`db.ts:318`) and the chain ends at `.version(15)` (`db.ts:451`, DA-650). A rebuild is `.version(16)` and also has to cover the `bookmark` table added in `.version(13)`, which did not exist when this was written.

**Diff review, for whoever rebuilds DA-182:**

- Running this branch's `DanmakuService.ts` through today's Biome gives `Found 2 errors. Found 47 warnings.` The added lines carry 4 `@ts-ignore` and 52 `any` tokens. (The PR's own `quality` check is red; its log is expired, `HTTP 410`, so I reproduced it locally instead.)
- `gunzipComments` does `catch (_e) { return [] }`. A bad blob renders an episode with zero comments and surfaces no error; the row still looks present. `gzipComments` mirrors it by silently writing an uncompressed blob on failure, which makes the on-disk format depend on `Blob.type` alone. `@danmaku-anywhere/result` exists for this.
- Nothing outside `DanmakuService` learns about the new table. On master `DataManagementService.ts:21` does `await this.db.episode.clear()` in a transaction that would not list `danmakuData`, so "clear data" would orphan the entire comment corpus and invert the storage win.
- No tests, in a repo that has had real migration test infra since `f6de2210 [DA-530]` (`src/common/db/db.migration.test.ts` drives the actual upgrade callback under `fake-indexeddb`).

**One risk I could not settle.** The migration and every write path `await` a `CompressionStream` round trip inside a Dexie `rw` transaction, and use neither `Dexie.waitFor()` nor any equivalent, even though dexie ships it precisely to hold a transaction open across non-IndexedDB work. I probed it rather than asserting it: a throwaway vitest running this PR's exact v12 upgrade against `fake-indexeddb` with 50 seeded episodes migrated all 50 cleanly (`{openError: null, migrated: 50, stillInline: 0, blobCount: 50}`), and a `Dexie.waitFor` variant was identical. So it does not reproduce there, and I am not claiming it as a bug. The takeaway is the inverse: `fake-indexeddb` does not enforce IDB auto-commit, so the migration test harness cannot catch transaction-lifetime bugs, and a DA-182 rebuild that compresses inside a transaction needs a real-browser check.

**Suggested shape for the rebuild:** `.version(16)` adding `danmakuData`; comment access confined behind `persistence/DanmakuService` on `this.db`; `wipeAllData` and the season cascade extended to the new table; `filterLite` / `filterCustomLite` changed to skip the blob read entirely (that is where the real win is); `Result`-typed decode failures; `Dexie.waitFor` around the compression; a `db.migration.test.ts` case for v15 to v16; and an e2e migration spec for what `fake-indexeddb` cannot cover.

PR #176, opened 2025-09-06
(extension) danmaku density chart [DA-201]

Mr-Quin:


Mr-Quin:
Task linked: [DA-201 Show danmaku density plot](https://app.clickup.com/t/90131020449/DA-201)

PR #200, opened 2025-11-26
(extension) use inversifyjs for di [DA-241]

Mr-Quin:


Mr-Quin:
Task linked: [DA-241 Use inversify js for DI](https://app.clickup.com/t/90131020449/DA-241)

PR #228, opened 2025-12-12
(extension) make tencent and bilibili providers work in automatic mode [DA-279]

Mr-Quin:
#185


Mr-Quin:
Task linked: [DA-279 Make Tencent and Bilibili providers work in automatic mode](https://app.clickup.com/t/90131020449/DA-279)

PR #301, opened 2026-02-02
(extension) make session DNR only apply to requests coming from the extension [DA-371]

Mr-Quin:


Mr-Quin:
Task linked: [DA-371 Make session DNR only apply to requests coming from the extension](https://app.clickup.com/t/90131020449/DA-371)

PR #323, opened 2026-03-15
(app) make local player page SSG [DA-396]

Mr-Quin:


Mr-Quin:
Task linked: [DA-396 SSG local player](https://app.clickup.com/t/90131020449/DA-396)

PR #385, opened 2026-05-14
(extension) suppress release notes auto-show in dev mode [DA-493]

Mr-Quin:
## Summary
- Guard the `showReleaseNotes: true` write in `OptionsManager.onInstalled` behind `!import.meta.env.DEV` so HMR/CRX rebuilds in dev no longer constantly re-pop the release notes modal.
- Production behavior unchanged — `import.meta.env.DEV` is compile-time replaced by Vite, so the guard disappears in shipped builds.
- The modal and query remain wired up; devs can still test by flipping `showReleaseNotes` manually via the BG service worker console.

## Verification
- `pnpm type-check`, `pnpm lint` clean.
- Not exercised via dev browser — the install/update listener doesn't fire during HMR sessions, so behavioral verification needs an uninstall/reinstall cycle. Please verify on next packaged update.

### DIFF
--- a/packages/danmaku-anywhere/src/background/syncOptions/OptionsManager.ts
+++ b/packages/danmaku-anywhere/src/background/syncOptions/OptionsManager.ts
@@ -27,7 +27,9 @@ export class OptionsManager {
       this.tryUpgradeOptions()
 
       if (details.reason === 'update') {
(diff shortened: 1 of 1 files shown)


Mr-Quin:
Task linked: [DA-493 Suppress release notes auto-show in dev mode](https://app.clickup.com/t/90131020449/DA-493)

PR #402, opened 2026-05-16
(extension) batch episodeFilterLite per season [DA-508]

Mr-Quin:
## Summary
- Rendering a season with ~260 episodes hung the list in the loading skeleton state because each `EpisodeRow` instantiated a `useQuery` that fired `chromeRpcClient.episodeFilterLite({ provider, indexedId, seasonId })` — 260 concurrent `chrome.runtime.sendMessage` roundtrips through the background, each doing its own IndexedDB lookup + season join.
- Replaced the per-row RPC with a single seasonId-scoped query at `EpisodeListInner` (`filterLite({ seasonId })`). Rows are now presentational: they receive their `EpisodeLite` (or `null`) by `indexedId` lookup from a memoized `Map`, plus a shared `isLoading`.
- Query semantics preserved: `refetchOnMount: false`, `refetchOnWindowFocus: false`, `staleTime: Infinity`.

## Test plan
- [x] `pnpm type-check`
- [x] `pnpm lint`
- [x] `pnpm --filter @mr-quin/danmaku-anywhere test` (273 passed)
- [ ] Manual: load a season with ~260 episodes in the dev browser, confirm the list renders without a long skeleton hang and the background log no longer spams `episodeFilterLite`

### DIFF
--- a/packages/danmaku-anywhere/src/common/components/EpisodeList/EpisodeSearchList.tsx
+++ b/packages/danmaku-anywhere/src/common/components/EpisodeList/EpisodeSearchList.tsx
@@ -1,5 +1,6 @@
 import type {
   CustomSeason,
+  EpisodeLite,
   EpisodeMeta,
   Season,
   WithSeason,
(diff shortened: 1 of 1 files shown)


Mr-Quin:
Task linked: [DA-508 Batch episodeFilterLite per season to fix loading hang on large episode lists](https://app.clickup.com/t/90131020449/DA-508)

PR #412, opened 2026-05-21
(extension) fix rem font-size regression on non-16px host roots [DA-522]

Mr-Quin:
## Summary
- The Sakura theme redesign hard-coded `rem` strings in typography variants and component overrides, bypassing MUI's `pxToRem`. On host pages with a non-default `html` font-size (e.g. Plex at 10px, or a host that uses 100px), text rendered at the wrong size because literal `rem` values resolve against the host page's `:root`.
- `createSakuraTheme` was also building the theme such that MUI's internal `pxToRem` closure stayed bound to the default `16`. `htmlFontSize` must be set on the first `createTheme` arg, not deep-merged in from `extra`, otherwise per-size styles in Button/Tab/ToggleButton/etc. compute against the wrong root.
- Swept consumer-side inline `rem` in `styled()` callbacks and `sx` props, preferring typography variants when possible.

## Test plan
- [ ] Reload extension; open popup. Buttons, chips, tabs, helper text render at intended sizes.
- [ ] Open the controller overlay on a host page with a non-16px `html` font-size (e.g. Plex). Verify FAB icon, floating panel, and all dialogs/forms render correctly.
- [ ] Verify both light and dark color modes.

### DIFF
--- a/packages/danmaku-anywhere/src/common/components/DanmakuFilter/RulesList.tsx
+++ b/packages/danmaku-anywhere/src/common/components/DanmakuFilter/RulesList.tsx
@@ -64,7 +64,7 @@ export function RulesList({
                 label={rules.length}
                 size="small"
                 variant="outlined"
-                sx={{ height: 20, fontSize: '0.6875rem' }}
+                sx={{ height: 20 }}
               />
             </>
           )}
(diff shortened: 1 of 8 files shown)


Mr-Quin:
Task linked: [DA-522 Fix rem font-size regression: extension UI scales when host html font-size is non-default](https://app.clickup.com/t/90131020449/DA-522)

PR #422, opened 2026-05-23
(chore) reverse gateway-crypto WASM to pure js [DA-533]

Mr-Quin:
## Summary

- Adds `packages/dango/src/helpers/gateway-decrypt.ts`: pure JS reimplementation of an upstream `gateway-crypto` WASM blob's `gateway_decrypt` export, so the upcoming source manifest that needs it can run without bundling a third-party binary
- Registers `$gatewayDecrypt(ciphertextB64, keyB64)` in the dango helper namespace
- Pins cipher equivalence to the upstream via captured fixture pairs

## Algorithm (extracted from wat disassembly)

- **Cipher**: AES-256-ECB with PKCS#7 padding
- **Key derivation**: base64-decode to 32 bytes → map each nibble through a fixed 16-entry permutation → 32-byte AES-256 key
- **ECB synthesis**: Web Crypto exposes only AES-CBC, so we append a manufactured fixup block (`AES_enc(0x10..0x10 XOR C_n-1)`) so subtle's mandatory PKCS#7 unpad succeeds, then un-chain in place. Strip the real PKCS#7 pad ourselves with a range check (catches wrong-key / corrupted-ciphertext inputs that would otherwise return garbage).

## Key configurability

The base64 key string is passed in by the manifest, not baked into the helper. Key rotation is a manifest edit, not a code change. The nibble permutation table stays baked in because (a) it hasn't rotated in known historical bundles, and (b) parameterizing it would require a per-call validity check that's noise for the realistic threat model.

## Verification

- 5 cipher fixtures (empty/single-byte/single-block/multi-block/UTF-8/JSON-shaped) pinned in `gateway-decrypt.test.ts`
- Negative cases: ciphertext that isn't a multiple of 16 bytes, key that doesn't decode to 32 bytes
- Smoke-tested against the live WASM with 5 fresh plaintexts; all round-tripped byte-identical

## Test plan

- [x] `pnpm --filter @danmaku-anywhere/dango test` (63 tests pass)
- [x] `pnpm lint` clean
- [ ] Reviewer to confirm the `extractable: false` flag on the imported CryptoKey is appropriate for this threat model
- [ ] Follow-up branch wires this into the manifest that needs it

## Risks

- Silent breakage if the upstream rotates the embedded key (and the manifest editor isn't updated). The helper throws a clear error if `keyB64` doesn't decode to 32 bytes; wrong-but-still-32-byte keys produce a PKCS#7 pad-length error rather than silent garbage.
- No live API call test (cipher equivalence is proven against WASM fixtures; since the upstream server runs the same WASM, this is sufficient).

### DIFF
--- a/packages/dango/AGENTS.md
+++ b/packages/dango/AGENTS.md
@@ -50,7 +50,7 @@ A `Manifest` is JSON validated by zod. Each manifest declares up to three `Pipel
 
 `src/helpers/registry.ts` is the closed namespace callable from JSONata as `$<name>`. Current set:
 
-- crypto: `$md5`
+- crypto: `$md5`, `$gatewayDecrypt`
 - codec: `$base64Encode`, `$base64Decode`, `$hexToInt`, `$bytesToHex`
 - text: `$regexExtract`, `$jsonParse`, `$jsonpUnwrap`, `$timeToSeconds`
 - query/signing: `$sortedQueryString`, `$sortedRawString`
--- a/packages/dango/src/__tests__/gateway-decrypt.test.ts
+++ b/packages/dango/src/__tests__/gateway-decrypt.test.ts
@@ -1,0 +1,56 @@
+import { describe, expect, it } from 'vitest'
+import { gatewayDecrypt } from '../helpers/gateway-decrypt.js'
+
+/**
+ * Pins gatewayDecrypt to byte-identical output of the upstream gateway-crypto
+ * reference. Fixtures were captured by running the upstream encrypt on each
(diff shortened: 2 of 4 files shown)


Mr-Quin:
Task linked: [DA-533 Reverse Migu gateway-crypto WASM to pure JS](https://app.clickup.com/t/90131020449/DA-533)

Mr-Quin:
Closing for now — the helper is dead code without the source manifest that consumes it. Will reopen (or fold into the manifest PR) once builtin-migu.json is ready. Branch DA-533_migu_wasm_reverse retained.

PR #430, opened 2026-05-25
(extension) in-video info panel [DA-543]

Mr-Quin:
## Summary

Phase 1 of the in-video **info panel**: a status surface mounted in the player shadow root that shows several feature statuses at once and is structured to grow into a general status/toast surface. Built as a multi-source, level-triggered model.

### Architecture
- **Source-keyed store.** The player store holds `entries: Partial<Record<PanelSource, PanelEntry>>`; `resolvePanelEntries` returns the present entries in a fixed order, one row per source. The panel is source-agnostic chrome (position, drag, idle auto-hide, PiP gate) around per-source rows: a shared header via `entryView` plus a body switched on source.
- **Two sources, two delivery paths.**
  - `pipeline` (danmaku status) is derived in the controller frame (`selectPanelState -> PipelineEntry | null`) and pushed over the existing `relay:command:syncPanelState` RPC.
  - `occlusion` is **player-local**: derived from `OcclusionService` signals and written straight to the store with no relay traffic. This is the case the keyed store exists for.
- **Display policy lives in the producer.** `selectPanelState` returns `null` when disabled or in manual mode before a successful mount, so `enabled`/`isManual` leave the wire and the player leaf carries no visibility policy.
- **Pull-on-connect resync.** The push is best-effort; the player pulls the current pipeline entry once its full server is listening (`relay:event:requestPanelState`), so a push dropped during the handshake does not leave a frame stuck on a stale terminal state.

### Occlusion source
- Added an `onRunningChange` signal to `OcclusionService` (the steady on/off state has no status reason) routed through `DanmakuManager`.
- `OcclusionEntryDeriver` folds the status reasons and the running flag into the entry: `downloading` is loading; `init`/`webgpu`/`taint`/`unavailable`/`segment` are errors (verified against the service: `init` means init-failed); running wins so an error only shows once the loop stops; a stopped loop with no error clears the row.

### UI
- Multi-row Card panel reusing the Sakura `--da-*` token language: per-row status dot (with pulse for in-progress), headline, collapsed glance vs panel-level expand, media block + count + localized provider chip, occlusion message line. Stable DOM contract (`data-da-ip-row`/`data-source`/`data-sev`, `data-da-ip-headline`/`-count`/`-source`/`-message`/`-tag`/`-grip`).
- New `infoPanel.enabled` extension option (default on); storage migrates to v27. Occlusion headline strings added (en/zh).

## Test plan
- [x] `pnpm type-check` and `pnpm lint` green
- [x] `pnpm test` — 646 unit tests pass (resolve, selectPanelState null-gate + priority + payload, occlusion deriver, idle tracker, sakura tokens)
- [x] `pnpm i18n:check` reports no drift
- [x] e2e `info-panel.spec.ts` passes locally: the pipeline row renders with its `data-source` + row-level `success` severity, the localized MacCMS chip, and the count on hover (real-browser DOM)
- [ ] Manual: occlusion row when segmentation runs (needs WebGPU + a downloaded model; unit-verified)

## Notes for reviewers
- Level-triggered only by design: the panel reflects current state; transient/auto-expiring toasts and an event/TTL channel are deferred. A third level-triggered source slots into the keyed store with no reshape; transient toasts (multiple, ordered, expiring) will need an additive structure beside it, not inside setEntry.
- The messaging here (push + pull-on-connect) is pragmatic; a follow-up will introduce a state-replication primitive that generalizes the resync. Tracked separately.


Mr-Quin:
Task linked: [DA-543 Info panel implementation](https://app.clickup.com/t/90131020449/DA-543)

Mr-Quin:
Reviewed all bot comments. Fixed in fe111f97:

- **Theme-aware colors** (light-mode readability): hardcoded `rgba(255,255,255,…)` → `color-mix(var(--da-text) N%)` at matching opacity (identical in dark mode, consistent with `SkipButton`).
- **`applyOffset`** now skips no-op updates (avoids redundant re-renders from the `ResizeObserver`).
- **Zero-size guard**: skip repositioning while the player is collapsed to 0×0 so the stored position isn't clobbered.
- **Docked tab `onClick`** for keyboard restore.
- **`PlayerIdleService`** drops the detached video on `videoNodeRemove` (fixes the detached-element retention).
- **`useWindowIdle` → `useWindowActive`** (matches its return value).
- **Settings** `createUpdate` spreads existing `infoPanel`.
- **e2e** text matchers anchored with `^…$`.

Not applied (not relevant to current HEAD), resolving:

- **`offsetParent` returns null in shadow DOM**: the panel mounts in light DOM under the positioned `uiLayer` (same tree), so `offsetParent` resolves correctly — confirmed by the passing positioning/visibility e2e.
- **`PlayerIdleService.start()` idempotency**: already guarded (early return when the tracker exists).
- **`InfoPanelService.mount()` re-attach**: `uiLayer` is a persistent singleton the panel rides along with; it isn't detached on video/parent swaps.
- **`usePersistedPosition` key reactivity**: the key is static at every call site and the consumer reads the initial value into a ref once.
- **Per-drag-frame layout read**: micro-opt on a tiny element; the new `applyOffset` guard already trims churn.
- **Density chart `else`/hide**: the chart defaults to hidden (no `--visible` class), so the branch is redundant.
- **`PanelStateBroadcaster` `commentCount`/`options.infoPanel`**: stale — the broadcaster no longer reads `options.infoPanel`, and `commentCount` is now used by the mounted state.

PR #435, opened 2026-05-29
(extension) render danmaku behind people via on-device segmentation [DA-9999]

Mr-Quin:
## Summary

Adds an opt-in **"render behind people"** mode: the playing video is segmented on-device (MediaPipe selfie segmenter) and the danmaku overlay is masked so comments render *behind* people instead of over them. Also lands a standalone video test harness shared by agentic dev iteration and Playwright e2e (replacing the `app/web` debug playground).

## How it works (the interesting part)

MediaPipe can't run in the player content script: its wasm loader injects a `<script>` into the page DOM, which executes in the **main world**, so the Emscripten factory registers on the wrong `window` (isolated/main world split, `ModuleFactory not set`). A content-script `Worker` is also blocked (cross-origin worker script).

So segmentation runs in a **hidden `chrome-extension://` iframe** (`pages/segmenter.html`): an extension-origin, single-world document where the loader works under the extension CSP even on strict-CSP hosts like YouTube. Assets are **bundled locally** (MV3 forbids remote code) and copied from `node_modules` at `(pre)build`/`(pre)dev`; the `extension_pages` CSP gains `wasm-unsafe-eval`.

- `OcclusionMaskService` (content script): a throttled `requestVideoFrameCallback` capture loop transfers frames to the iframe, `maskGeometry` builds a letterbox-correct alpha mask, and it is applied via the `@mr-quin/danmu` `updateOccludedUrl` hook. The danmaku render path is unchanged; the iframe is a sidecar, only a CSS mask URL touches the overlay.
- Gated behind a `danmakuOptions` flag (default **off**, Chrome only). Cross-origin-without-CORS and DRM/EME video taint the canvas and disable the feature for that video.

## Dev / e2e harness

- Standalone pages under `e2e/fixtures/sites` + a dependency-free `e2e/harness/serve.mjs` (HTTP Range) + checked-in local videos. `pnpm dev:harness` serves it; the `app/web` debug playground is removed.
- Deterministic occlusion e2e via a mock mask provider (no ML/network) plus a Range media route; the real model never runs in CI.

## Verification

- `pnpm lint` (tsc + biome) clean; **323 unit tests**; **full e2e suite green (36 specs)** incl. the new occlusion specs (enabled/disabled), auto-mount, migration, mount, lifecycle, and sources; `app/web` type-check clean.
- Manually verified in a real browser incl. a real-person clip: the segmenter runs on the GPU and danmaku render cleanly behind a person.

## Notes

- `DA-9999` is a placeholder id (ClickUp not configured in this environment); retag as needed.
- `selfie_segmenter.tflite` (~244KB) is committed; the ~18MB wasm is gitignored and regenerated from `node_modules` at build.


Mr-Quin:
image

Mr-Quin:
Good god claude
image

Mr-Quin:
Superseded by #437 (DA-554), the clean people-occlusion infra PR off latest master. The prototype branch is kept for reference for the anime/hosting follow-ups (DA-555/556).

PR #439, opened 2026-05-31
(extension) occlusion model abstraction, on-device cache, and anime ORT provider [DA-550]

Mr-Quin:
## Summary
- Adds a model-runtime abstraction behind the existing default-off occlusion feature. A registry maps the `occlusionModel` option (`people | anime`) to a runtime (`mediapipe | ort`), inference size, and (for hosted models) a download URL + sha256; `maskProviderFactory` resolves the runtime and the segmenter iframe dispatches on it.
- Occlusion runtime is an injectable service (`OcclusionService`, `.service.ts`) that swaps providers via `configure()`; `DanmakuManager` injects it. The e2e/mock provider is selected at the DI container level (`uiIoc`), not inside the factory.
- Anime path: ISNet via `onnxruntime-web` on WebGPU, lazily imported so ORT stays off the default path. Frames are letterboxed (aspect-preserving) into the square input, normalized `/255`, and the model's `[0,1]` foreground alpha is thresholded directly; a degenerate-mask guard suppresses frames that would hide every danmaku. Output is read from the model's own tensor dims with a shape guard.
- Model delivery: a generic `opfsFileCache` (`fetchAndCacheFile`) fetches the hosted model with sha256 integrity, streams progress, and caches it in OPFS. Verified the cache is global across host sites (extension frames are exempt from Chrome's top-level-site partitioning), so it downloads once and takes one disk copy, not per site. The model is not bundled; the fetch is content-hash cache-busted so a re-upload is never served stale from a CDN edge.
- Runtime wasm (mediapipe + ort) is vendored from `node_modules` into `public/` by a vite plugin (`vendorRuntimeAssets`) at `buildStart`, replacing the `package.json` pre-build/pre-dev copy scripts.
- Download UX: progress is reported from the iframe; the first chunk shows a one-time "downloading" info toast, and the init timeout becomes an inactivity watchdog so a large first download does not trip it.
- DevTools "Segmentation" panel: live running/model/fps/last-error plus a debug-overlay toggle.
- Occlusion failures are classified (`init | taint | webgpu | segment | unavailable`) and surfaced as a toast instead of failing silently; playback is never interrupted.
- Settings gain a People/Anime model picker. en/zh i18n. No new migration (`occlusionModel` already shipped).

## Test plan
- [ ] `pnpm --filter @mr-quin/danmaku-anywhere test` (unit; occlusion registry/cache/service/geometry)
- [ ] `VITE_DA_ENV=e2e pnpm --filter @mr-quin/danmaku-anywhere build && pnpm --filter @mr-quin/danmaku-anywhere test:e2e e2e/specs/integration/occlusion.spec.ts e2e/specs/integration/occlusion-cache-scope.spec.ts` — non-theatre failure test (anime without WebGPU surfaces a visible error toast and applies no mask) plus a guard that the model cache OPFS is global across top-level sites.
- [ ] Manual on a WebGPU browser: enable occlusion, switch People/Anime, confirm the download toast, the anime mask renders behind danmaku, and the Segmentation devtools panel shows live stats.

## Notes
- **Hosted anime model.** The anime model is the 320-input ISNet export hosted at `https://assets.danmaku.weeblify.app/models/anime-isnet.onnx` (sha256 `bed52e3bc068689c1b37f73871e184ae65e183cb6232097bb7fb72a72372f60e`, 176166467 bytes). The fetch verifies this hash; a re-host must match it. The file is gitignored and not in the repo. The people path is independent and unaffected if the anime model is missing.
- Anime preprocessing was root-caused and corrected (letterbox + `/255` + direct `[0,1]` alpha, not the earlier min-max/stretch) and the model re-exported at 320 for ~9x faster inference at equal quality; verified visually on WebGPU hardware. e2e keeps the real model out of CI via an engine-aware mock.
- **Model-management UI is minimal:** a download notice toast only. A fuller manager and a declarative per-model manifest to support arbitrary ORT segmentation models (and, later, user-supplied models) are tracked as a follow-up subtask under the occlusion epic.
- Minor, debug-panel-only: the overlay toggle and idle `debugOverlay` stat can briefly diverge from runtime across a quality/model change; the error-reason classification uses a message heuristic.


Mr-Quin:
Task linked: [DA-550 Model wiring: provider / engine / registry abstraction](https://app.clickup.com/t/90131020449/DA-550)

Mr-Quin:
Task linked: [DA-551 Model hosting (R2 + OPFS cache, no bundling) + model-management UI](https://app.clickup.com/t/90131020449/DA-551)

packages/danmaku-anywhere/src/content/controller/danmaku/frame/FrameManager.tsx
```diff
@@ -96,6 +99,40 @@ export const FrameManager = () => {
     reEvaluateActiveFrame()
   })
 
+  const handleOcclusionStatus = useEventCallback((status: OcclusionStatus) => {
```
Mr-Quin, line 136: these should be outside of the component. we have a pattern using i18n.t

packages/danmaku-anywhere/src/content/player/occlusion/modelCache.ts
```diff
@@ -0,0 +1,278 @@
+/**
```
Mr-Quin: comments way too verbose.

packages/danmaku-anywhere/src/content/player/occlusion/createMaskProvider.ts
```diff
@@ -1,14 +1,16 @@
...
 import { MockMaskProvider } from './MockMaskProvider'
+import { getModel } from './modelRegistry'
 import type { MaskProvider } from './types'
 
-export function createMaskProvider(): MaskProvider {
+export function createMaskProvider(model: OcclusionModel): MaskProvider {
   // e2e runs a deterministic mock so occlusion specs assert our pipeline
   // (capture, geometry, compositing) rather than the non-deterministic ML
   // model or a bundled person video.
   if (IS_DA_E2E) {
```
Mr-Quin: why not an injectable

packages/danmaku-anywhere/src/common/storage/opfsFileCache.ts
```diff
@@ -0,0 +1,278 @@
...
+  onProgress?: (progress: ModelCacheProgress) => void
+  /** Cancels an in-flight fetch (does not affect an OPFS cache hit). */
+  signal?: AbortSignal
+}
+
+/**
+ * Minimal OPFS surface used by the cache. The default implementation wraps
+ * navigator.storage; tests inject a fake.
+ */
+export interface OpfsAdapter {
```
Mr-Quin, line 18: opfs layer should be generic and be usable to cache any large file

PR #446, opened 2026-06-03
(proxy) DDP proxy: additive transparent /api passthrough [DA-574]

Mr-Quin:
## What

Adds a transparent `/ddp/api/*` passthrough in `backend/proxy` that forwards to the `DDP_SERVICE` microservice (strips the `/ddp` mount prefix, no `?path=` rewrite), preserving the existing `da-authenticated` handling. The legacy `/ddp/v1?path=` route is **unchanged** — purely additive.

Pairs with `Mr-Quin/ddp-microservice#4` (the matching transparent `/api/*` route in the microservice).

## Why

dango's `builtin:dandanplay` is being unified into a single manifest that targets the official DDP shape (`{baseUrl}/api/v2/...`) directly. The extension's built-in DDP config will set `baseUrl` to this proxy, so we need a transparent endpoint here; the proxy/microservice keep holding the credentials so the client carries none.

## Tests / verification

- New unit tests: transparent `/api/*` forwards the stripped path to `DDP_SERVICE`; sets `da-authenticated` when signed in. Full `danDanPlay` suite: 8 pass.
- Deployed to staging and verified live: `https://api.danmaku-staging.weeblify.app/ddp/api/v2/search/anime?keyword=frieren` returns real DanDanPlay results; `/ddp/v1?path=` still works.

> No ClickUp task linked yet (opened directly per request); happy to attach one.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/backend/proxy/src/routes/api/ddp/danDanPlay.test.ts
+++ b/backend/proxy/src/routes/api/ddp/danDanPlay.test.ts
@@ -129,6 +129,20 @@ describe('DanDanPlay API', () => {
     )
   })
 
+  it('transparent route strips client-supplied da-authenticated header when not signed in', async () => {
+    const app = createAppWithUser(null)
+    const request = new Request(
+      createTestUrl('/ddp/api/v2/search/anime', {
+        query: { keyword: 'nichijou' },
+      }),
+      { headers: { 'da-authenticated': '1' } }
+    )
+    await makeUnitTestRequest(request, { app })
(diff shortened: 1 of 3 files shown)


Mr-Quin:
Task linked: [DA-574 DDP proxy: additive transparent /api passthrough](https://app.clickup.com/t/90131020449/DA-574)

PR #452, opened 2026-06-05
(proxy) serve dango manifest catalog + files [DA-568]

Mr-Quin:
## Summary

Adds a `/api/manifest` router to the proxy backend, mirroring the existing `/api/kazumi/rules` proxy pattern. It lets clients fetch the dango manifest catalog and individual manifest files through the backend rather than hitting GitHub raw directly.

- `GET /api/manifest/` proxies `https://raw.githubusercontent.com/Mr-Quin/dango/main/packages/dango-manifests/catalog.json`. Typed via hono-openapi (`describeRoute` + `resolver` + zod): `{ packageVersion: string, manifests: Array<{ id, name, version, apiVersion, file }> }`.
- `GET /api/manifest/file?file=<path>` validates `file` against `^src/manifests/[\w.-]+\.json$` and proxies `.../dango-manifests/${file}`. The regex rejects path traversal, absolute URLs, and non-manifest paths (HTTP 400 via the zod query validator), preventing arbitrary-fetch/SSRF.
- Both routes use `useCache({ maxAge: 3600 })` (1 hour TTL) since manifests change rarely.
- Mounted at `/manifest` in `routes.ts` next to `/kazumi`.

### Conservative choices (ambiguous points)

- Used a single `router.ts` mounted directly at `/manifest` rather than the kazumi-style nested `router.ts` + `rules.ts`. Kazumi nests because `rules` is a sub-resource (`/kazumi/rules`); the manifest endpoints live directly under `/manifest`, so an extra nesting layer would be noise. The implementation otherwise mirrors kazumi exactly.
- The `/file` response schema is left as a permissive passthrough object since manifest shape is owned by the dango repo and not duplicated here.

## Test plan

- New `router.test.ts` (6 cases, mirrors `kazumi/rules.test.ts`): asserts the catalog and file routes call `fetch` with the correct upstream dango URLs, and that the `file` validator returns 400 for missing param, path traversal (`../../../../etc/passwd.json`), non-manifest paths (`catalog.json`), and absolute URLs (`https://evil.example.com/...`) without ever calling `fetch`.
- Verified upstream `catalog.json` returns HTTP 200.
- `biome check` clean on the new files and `routes.ts`.
- Full proxy vitest suite: 66 passed (13 files).

Note: `tsgo` type-check could not run in the local worktree due to a pre-existing `tsconfig.json` `baseUrl` incompatibility with the worktree's installed tsgo dev build (errors are repo-wide, not in the new code); CI runs the pinned version.

### DIFF
--- a/backend/proxy/src/routes/api/manifest/router.test.ts
+++ b/backend/proxy/src/routes/api/manifest/router.test.ts
@@ -1,0 +1,120 @@
+import { afterEach, describe, expect, it, vi } from 'vitest'
+import { createTestUrl } from '@/test-utils/createTestUrl'
+import { makeUnitTestRequest } from '@/test-utils/makeUnitTestRequest'
+
+const dangoBaseUrl =
+  'https://raw.githubusercontent.com/Mr-Quin/dango/main/packages/dango-manifests'
+
+describe('Manifest API', () => {
+  afterEach(() => {
+    vi.restoreAllMocks()
+  })
+
+  it('proxies the catalog from the dango repo (GET /)', async () => {
+    const mockCatalog = {
+      packageVersion: '0.2.0',
+      manifests: [
+        {
+          id: 'builtin:bilibili',
+          name: 'Bilibili',
+          version: '0.2.0',
+          apiVersion: 1,
+          file: 'src/manifests/builtin-bilibili.json',
+        },
+      ],
+    }
+
+    const fetchSpy = vi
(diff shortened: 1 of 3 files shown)


Mr-Quin:
Task linked: [DA-568 dango: remote/CDN manifest distribution channel (Phase 5)](https://app.clickup.com/t/90131020449/DA-568)

PR #453, opened 2026-06-05
(extension) storage-backed dynamic manifest registry [DA-482]

Mr-Quin:
## Summary

Converts `ManifestRegistry` from static JSON imports into a **storage-backed, async-initialized** registry, without changing which manifests register or any user-facing behavior. This is the foundation for later catalog import/update (catalog fetching, import UI, and id-rename/migration are intentionally out of scope).

- New `ManifestStore` (`chrome.storage.local`, single `manifests` key) holding `Record<manifestId, { manifest, kind: 'preinstalled' | 'user' }>`. Methods: `getAll`, `get`, `has`, `set`, `setMany`, `remove`. An `IManifestStore` interface is extracted for testability/injection.
- `ManifestRegistry` now reads the store on init. If empty (fresh install / upgrade) it **seeds from the currently bundled manifest JSON imports** with `kind: 'preinstalled'` and persists. It builds one `ManifestRunner` per stored manifest via `zManifest.safeParse`, skipping + logging failures (apiVersion-incompatible) instead of crashing.
- New methods: `register(manifest, kind)`, `unregister(id)`, `list()`, `refresh()`. `getRunner(id)` stays synchronous and throws if missing, as before.
- Async-init ordering: the registry exposes a `ready` promise (kicked off in the constructor, mirroring `ReadinessService`). It is warmed eagerly at background startup and awaited by consumers before any `getRunner` call.

Behavior is preserved: the same builtin manifests (dandanplay, bilibili, tencent) register after this change, seeded from the bundle.

### Consumer call sites that needed an async/await change

- `ManifestProviderService`: `search`, `getSeason`, `getEpisodes`, `getDanmaku`, `parseUrl` now `await this.registry.ready` before resolving a runner (all already async).
- `ProviderService`: `parseUrl` (awaits before the canParse loop, the only sync `getRunner` path), `probeLogin`, and `getManifestSpec` await readiness. **`getManifestSpec` changed from sync to `async`** (it was unused, so no callers were affected).
- `src/background/index.ts`: warms the registry at startup via `void container.get(ManifestRegistry).ready`.

## Test plan

- New `ManifestStore.test.ts`: empty `getAll` default, `get`/`has` resolution, `set` merge, `remove` (drops one entry, no-op when absent).
- New `ManifestRegistry.test.ts`: seeds bundled manifests as `preinstalled` when empty; does not reseed when already populated (idempotent init); `getRunner` returns seeded runners and throws on unknown id; `register`/`unregister` mutate both store and runner map; a `safeParse` failure (apiVersion 999) is skipped, not fatal.
- `pnpm --filter @mr-quin/danmaku-anywhere lint` (tsc + biome) passes on the touched files. Pre-existing biome warnings in unrelated files are untouched.
- Provider + matching vitest suites pass (30 tests).

### DIFF
--- a/packages/danmaku-anywhere/src/background/index.ts
+++ b/packages/danmaku-anywhere/src/background/index.ts
@@ -6,6 +6,7 @@ import { PortsManager } from '@/background/ports/PortsManager'
 import { RpcManager } from '@/background/rpc/RpcManager'
 import { MountConfigTabReloader } from '@/background/scripting/MountConfigTabReloader'
 import { ScriptingManager } from '@/background/scripting/ScriptingManager'
+import { ManifestRegistry } from '@/background/services/providers/ManifestRegistry'
 import { OptionsManager } from '@/background/syncOptions/OptionsManager'
 import { deferredConfigureStore } from '@/background/utils/deferredConfigureStore'
 import { generateId } from '@/background/utils/generateId'
@@ -32,6 +33,8 @@ container.get(RpcManager).setup()
 container.get(NetRequestManager).setup()
 container.get(AlarmManager).setup()
 container.get(PortsManager).setup()
+
+void container.get(ManifestRegistry).ready
 
 setLogService(container.get(LogService))
 
--- a/packages/danmaku-anywhere/src/background/services/providers/IDanmakuProvider.ts
+++ b/packages/danmaku-anywhere/src/background/services/providers/IDanmakuProvider.ts
@@ -40,7 +40,7 @@ export interface IDanmakuProvider {
     episodeNumber: number
   ): Promise<WithSeason<EpisodeMeta> | null>
 
-  canParse?(url: string): boolean
+  canParse?(url: string): Promise<boolean>
 
   parseUrl?(url: string): Promise<ParseUrlResult | null>
 }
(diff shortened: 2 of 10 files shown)


Mr-Quin:
Task linked: [DA-482 Generic ManifestProviderService + manifest registry](https://app.clickup.com/t/90131020449/DA-482)

Mr-Quin:
Addressed the review in `324c20f4`:

- **DI / startup crash (critical):** `ManifestStore` is now `@injectable`, and `ManifestRegistry` `@inject(ManifestStore)`s it. No more undecorated interface ctor param with a `new ManifestStore()` default (which would crash the autobind container at resolution).
- **Concurrent writes (high):** `ManifestStore` set/setMany/remove now serialize via `async-mutex`'s `Mutex` (the project's existing pattern, as in `setSessionHeader.ts`), so the read-modify-write on the single storage key can't clobber.
- **await ready in register/unregister/refresh (medium):** added (prior commit).
- **loadEntry keys runners by the store id (medium):** done.
- **Persist `parsed.data` instead of the raw manifest (medium):** declined. `zManifest` injects defaults (`protoDescriptors: {}`, `urlMatch: []`), so persisting parsed.data changes the stored shape vs the import. The runner is always built from validated `parsed.data` and the stored manifest is re-validated on every load, so the raw form is functionally equivalent and stays faithful to the imported source.

PR #461, opened 2026-06-06
(extension) assert DB provider ids migrate to bare in smoke test [DA-588]

Mr-Quin:
## Summary

Follow-up to #459. The migration smoke test verified season/episode/customEpisode **counts** after the cross-version swap, but never asserted that `season.providerConfigId` was actually remapped from `builtin:*` to bare by the Dexie v14 migration. So the v14 DB remap had no real coverage.

This adds, in the same v1.5.0 → current swap:
- **Before swap:** assert the seeded seasons carry at least one `builtin:*` provider id (v1.5.0's DB v12 backfills `season.providerConfigId` with `builtin:<x>`), so there is genuinely something for v14 to strip.
- **After swap:** assert no season retains a `builtin:` provider id, and every season's `providerConfigId` still resolves to an existing provider config (no orphaned reference).

## Verification

Migration smoke test passes on the current-master tree:
```
✓ migration swap › prior install, seed via UI, swap, upgrade preserves seeded data
```

Test-only change; no source changes.

### DIFF
--- a/packages/danmaku-anywhere/e2e/specs/migration/migration-smoke.spec.ts
+++ b/packages/danmaku-anywhere/e2e/specs/migration/migration-smoke.spec.ts
@@ -92,6 +92,7 @@ async function runSwap(tmpRoot: string): Promise<BrowserContext> {
   const seededProbe = await openProbePage(context)
   const seededSync = await readSyncSnapshot(seededProbe)
   const seededIdb = await readIdbCounts(seededProbe)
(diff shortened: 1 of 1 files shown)


Mr-Quin:
Task linked: [DA-588 Migrate extension to dango 0.3.0: drop builtin: prefix end-to-end (ids + stored data)](https://app.clickup.com/t/90131020449/DA-588)

PR #469, opened 2026-06-07
(app) load scraped videos CORS-clean for occlusion frame reads [DA-601]

Mr-Quin:
## Summary
- Set `crossorigin=anonymous` on the web app's Artplayer `<video>` (via `moreVideoAttr`) so the scraped video loads as a CORS request the occlusion feature can read frames from directly, no clone or second fetch.
- Extend the existing source-origin DNR header rule (`setRequestHeaderRule`) with an optional `origin`: when set, it rewrites the request `Origin` to the source origin and forces `Access-Control-Allow-Origin: *` on the response, so a hotlink-protected CDN that only allowlists the source origin still serves the CORS request origin-clean. Threaded from `kazumi.service` using the same `sourceOrigin` already used for `Referer`.
- No-cors fallback: on `video:error` while `crossorigin` is set, drop the attribute so Artplayer's reconnect reloads the video no-cors; playback is never broken (occlusion just stays off for that source). The expected, recoverable CORS error toast is suppressed once; genuine failures still surface.

## Test plan
- [ ] `pnpm --filter @danmaku-anywhere/web-scraper test` (covers the header-policy change: `Origin` + `Access-Control-Allow-Origin` added only when `origin` is provided, `Referer` preserved).
- [ ] On the CF preview: with occlusion enabled, play a real Kazumi/AGE source (agedm.io / toutiao CDN) and confirm the video loads CORS-clean (no 504) and the occlusion mask applies.
- Faithful e2e of the real hotlink CDN is not feasible (live third-party CDN), and this change is in the web app + web-scraper, not the extension Playwright suite, so no extension e2e spec is added. The deterministic header-policy logic is unit-tested; the player CORS/no-cors fallback is verified via the CF preview above.

## Notes
- The `Origin`/`ACAO` rewrite reuses the same broad rule as `Referer` (per design): injecting `Origin: sourceOrigin` toward a CDN that by premise allowlists that origin cannot be more restrictive than omitting it, and `ACAO: *` with `crossorigin=anonymous` exposes only anonymous public content (no credentials).

### DIFF
--- a/app/web/src/app/core/video-player/video.service.ts
+++ b/app/web/src/app/core/video-player/video.service.ts
@@ -199,8 +199,7 @@ export class VideoService {
       airplay: false,
       lang: 'zh-cn',
       theme: '#f472b6',
-      // load CORS-clean so occlusion can read frames from the video; falls back
-      // to no-cors on error (see setupEventListeners) for CDNs that reject it
+      // CORS-clean so occlusion can read frames; falls back to no-cors on error
       moreVideoAttr: { crossOrigin: 'anonymous' },
     })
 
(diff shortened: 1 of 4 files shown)


Mr-Quin:
Task linked: [DA-601 Web app: load scraped videos CORS-clean so occlusion can read frames](https://app.clickup.com/t/90131020449/DA-601)

PR #470, opened 2026-06-07
(extension) occlusion: close test-coverage gaps from cross-origin capture [DA-602]

Mr-Quin:
## Summary
- Add unit tests for `crossOriginCapture.ts` (previously zero coverage, the most-churned code in the occlusion feature). Mocks `document.createElement('video')`/`canvas` and stubs `chromeRpcClient` to exercise: the `isVideoOriginClean` origin probe (readyState guard, SecurityError vs other), clone `waitReady` (already-loaded sync kick, readyState-gated resolve, error, timeout), `sync` (playbackRate match, drift-triggered seek, pause/play mirroring), `dispose` idempotency, and the setup RPC-failure graceful path.
- Add a failure-path e2e to `occlusion-cross-origin.spec.ts`: a cross-origin source the DNR rule cannot rescue surfaces the taint status (error toast), keeps playback intact (playhead advances), and applies no mask. This is the first test of the "fails gracefully without breaking playback" guarantee.
- Extend the harness (`e2e/harness/serve.mjs`) with a `/cors-fail/` route that 403s the crossorigin clone's CORS request (it carries an `Origin` header) while the plain `<video>` (no `Origin`) still plays and taints, so the DNR ACAO rule genuinely cannot recover it.

## Test plan
- [ ] `pnpm --filter @mr-quin/danmaku-anywhere test` (covers `src/content/player/occlusion/crossOriginCapture.test.ts`)
- [ ] `pnpm --filter @mr-quin/danmaku-anywhere test:e2e e2e/specs/integration/occlusion-cross-origin.spec.ts` (the new "occlusion degrades safely when the tainted source cannot be recovered" case plus the existing recovery cases)
- [ ] `pnpm lint`

## Notes
- The crossOriginCapture unit tests are intended to migrate/extend into the FrameSource tests once the sibling refactor lands.

### DIFF
--- a/packages/danmaku-anywhere/e2e/harness/serve.mjs
+++ b/packages/danmaku-anywhere/e2e/harness/serve.mjs
@@ -89,14 +89,13 @@ const CORS_FAIL_PREFIX = '/cors-fail/'
 
 function handle(req, res) {
   let urlPath = req.url
-  const corsFail = urlPath.startsWith(CORS_FAIL_PREFIX)
-  if (corsFail) {
-    urlPath = `/${urlPath.slice(CORS_FAIL_PREFIX.length)}`
+  if (urlPath.startsWith(CORS_FAIL_PREFIX)) {
     if (req.headers.origin) {
       res.writeHead(403)
       res.end('Forbidden')
       return
     }
+    urlPath = `/${urlPath.slice(CORS_FAIL_PREFIX.length)}`
   }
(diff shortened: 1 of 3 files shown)


Mr-Quin:
Task linked: [DA-602 Occlusion: close test-coverage gaps (failure-path e2e + crossOriginCapture unit tests)](https://app.clickup.com/t/90131020449/DA-602)

PR #477, opened 2026-06-08
(chore) add release-manager dev utility for extension builds [DA-612]

Mr-Quin:
## Summary
- New dev-tooling package `tools/release-manager` (`@danmaku-anywhere/release-manager`, private). Adds a `tools/**` workspace glob.
- Local-web utility: a Node/Hono backend (full FS access, bound to `127.0.0.1`) serves a single-screen React UI to **download** extension release/preview builds from GitHub, **cache** them per tag, and **swap** which cached build is the active unpacked folder loaded into `chrome://extensions`. Solves the slow download/unpack/overwrite-and-reload loop for manual release testing.
- Shell-agnostic, `Result`-based core (`github` / `store` / `cache` / `active` / `manager`) with no HTTP/UI coupling, so a Tauri/Electron shell can wrap it later. Scope is folder-manager only (no Chrome launcher); Chrome only (Firefox deferred); dev-first (user-facing deferred).
- No `build` script by design, so the tool stays out of root `pnpm build`; it still joins `test` / `type-check` / `lint:ci` sweeps.

## Implementation notes
- Releases grouped Stable vs Preview by the API `prerelease` flag; preview subtype from literal tag prefixes (`nightly-`, `preview-pr-`, `preview-branch-`, `preview-manual-`). Chrome asset matched by `endsWith('-chrome.zip')`.
- Stable releases are published as drafts, so listing/downloading them needs a GitHub token; the optional token is threaded into both the releases fetch and the asset download (`Accept: application/octet-stream`).
- Downloads unzip (`fflate`) into a temp dir then atomically rename into `cache/<tag>/`; the builds index is written only on success and reconciled against on-disk dirs at startup.
- **Active folder is a real directory with the build copied in, not a symlink.** A sandboxed (Flatpak) Chrome reads the picked unpacked folder through a document portal that does not grant a symlink's target, so a symlinked active folder fails with "File path cannot be resolved." Copying keeps the loaded path stable across swaps and works for both native and Flatpak Chrome; the copy source is the local cache, so swapping never re-downloads. Verified end-to-end in a real (Flatpak) Chrome: load once, then Set active + reload swaps the build in place.
- Security posture: token stored in `config.json` at mode `0600`, never serialized to the browser (state exposes `hasToken` only); server is loopback-only.

## Test plan
- [ ] `pnpm --filter @danmaku-anywhere/release-manager test` (37 tests: github grouping/asset resolution, atomic unzip + index reconcile, active-folder copy/replace/clear, config round-trip + token masking + 0600, manager download-guard + active-build delete refusal + token threading)
- [ ] `pnpm --filter @danmaku-anywhere/release-manager type-check` and `lint`
- [ ] Manual: `pnpm release-manager`, download a build, Set active, "Load unpacked" `<dataDir>/active` in `chrome://extensions`, Set-active another build, reload the extension (contents swap in place).
- No extension e2e: this package is outside the Playwright harness and has no extension runtime surface.

## Notes
- Releases list shows the most recent 100 releases (no pagination), which is fine for a dev tool wanting recent builds; documented in `AGENTS.md`.

### DIFF
--- a/package.json
+++ b/package.json
@@ -16,6 +16,7 @@
     "dev:extension": "pnpm -r -F \"./packages/danmaku-anywhere\" dev",
     "dev:extension:standalone": "pnpm -r -F \"./packages/danmaku-anywhere\" standalone:dev",
     "build": "pnpm -r build",
+    "release-manager": "pnpm --filter @danmaku-anywhere/release-manager start",
     "build:packages": "pnpm -r -F \"./packages/**\" build",
     "format": "pnpm -r -F \"./packages/**\" format",
     "lint": "pnpm -r -F \"./packages/**\" lint",
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -143,7 +143,7 @@ importers:
     devDependencies:
       '@angular/build':
         specifier: ^21.0.0
-        version: 21.2.12(043cb7c17b8a83e84fe19acdc1f9a212)
+        version: 21.2.12(0b108dd98d30e1c139943813f67d6faf)
       '@angular/cli':
         specifier: ^21.0.0
         version: 21.2.12(@types/node@24.12.4)(chokidar@5.0.0)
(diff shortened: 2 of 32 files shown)


Mr-Quin:
Task linked: [DA-612 Dev utility to download and manage extension release builds](https://app.clickup.com/t/90131020449/DA-612)

PR #478, opened 2026-06-08
(chore) trim ML wasm from extension bundle [DA-615]

Mr-Quin:
## Summary
- Trim the vendored ML wasm that ships in the extension but is never loaded at runtime, cutting a fresh prod build from ~123M to ~54M.
- `vendorRuntimeAssets.ts`: onnxruntime-web group keeps only the asyncify pair (`ort-wasm-simd-threaded.asyncify.{wasm,mjs}`); drops the plain and jsep variants (~39M). MediaPipe group drops the no-SIMD fallback (`vision_wasm_nosimd_internal.{js,wasm}`, ~9M) that the chrome89+/firefox89+ build targets never select.
- `vite.config.ts`: add `onnxruntime-web-use-extern-wasm` to `resolve.conditions` (spread over `defaultClientConditions` so Vite's defaults are preserved) so `onnxruntime-web/webgpu` resolves to the extern-wasm build (`ort.webgpu.min.mjs`). The default bundled build inlines the wasm via `new URL`, making Vite emit a duplicate 23M asyncify wasm into `assets/` that the runtime `wasmPaths` override never loads.
- `vendorRuntimeAssets.ts`: wipe each vendored dir before copying, so a trimmed file list can't leave stale runtimes behind to ship (these dirs hold only vendored assets).

## Test plan
- [ ] `cd packages/danmaku-anywhere && pnpm build`, then `du -sh build build/ort build/mediapipe build/assets`. Expect total ~54M; `build/ort` holds only the asyncify pair; no `*.wasm` in `build/assets`; no `nosimd` in `build/mediapipe`.
- [ ] `pnpm --filter @mr-quin/danmaku-anywhere test:e2e -- e2e/specs/integration/occlusion.spec.ts e2e/specs/integration/occlusion-models.spec.ts e2e/specs/integration/occlusion-cross-origin.spec.ts e2e/specs/integration/occlusion-cache-scope.spec.ts` (9 tests, all pass against the trimmed build).

## Notes
- Verified the real segmenter runtime in a browser (the e2e suite cannot: CI has no WebGPU, the happy path uses a MockMaskProvider, and the model specs use fake OPFS blobs with no real download). Drove `pages/segmenter.html` directly:
  - MediaPipe selfie path initializes (`init ok`); network shows `vision_wasm_internal.{js,wasm}` + the tflite model at 200, and the SIMD build is auto-picked (no `nosimd` fetch).
  - ORT WebGPU path with the real hosted `anime` ISNet model initializes a session (`init ok`); network shows `ort.webgpu.min` chunk, the model, and `ort-wasm-simd-threaded.asyncify.{mjs,wasm}` at 200, with no `jsep`/plain fetch and no duplicate `assets/*.wasm`.
  - The only file-not-found entries were deliberate probes of the removed files, confirming they no longer ship.
- No new e2e coverage added: a missing-wasm regression only surfaces when the real runtime loads, which the suite intentionally avoids (no WebGPU/GPU in CI). The browser verification above covers that gap.

### DIFF
--- a/packages/danmaku-anywhere/scripts/vendorRuntimeAssets.ts
+++ b/packages/danmaku-anywhere/scripts/vendorRuntimeAssets.ts
@@ -32,22 +32,19 @@ const ASSET_GROUPS: readonly AssetGroup[] = [
         'wasm'
       ),
     destDir: 'mediapipe/wasm',
-    files: [
-      'vision_wasm_internal.js',
-      'vision_wasm_internal.wasm',
-      'vision_wasm_nosimd_internal.js',
-      'vision_wasm_nosimd_internal.wasm',
-    ],
+    // The build targets (chrome89+/firefox89+) all support SIMD, so
+    // FilesetResolver always picks the SIMD build; the no-SIMD fallback is dead
+    // weight.
+    files: ['vision_wasm_internal.js', 'vision_wasm_internal.wasm'],
   },
   {
     label: 'onnxruntime-web',
     resolveSrcDir: () => path.dirname(require.resolve('onnxruntime-web')),
     destDir: 'ort',
+    // The extern-wasm webgpu build (ort.webgpu.min.mjs, selected by the resolve
+    // condition in vite.config) loads only the asyncify pair via the wasmPaths
+    // override; the plain and jsep variants are never fetched.
     files: [
-      'ort-wasm-simd-threaded.wasm',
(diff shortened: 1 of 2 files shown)


Mr-Quin:
Task linked: [DA-615 Reduce extension bundle size (trim unused ORT/MediaPipe wasm)](https://app.clickup.com/t/90131020449/DA-615)

PR #482, opened 2026-06-08
(chore) capture recurring agent gotchas in committed docs [DA-620]

Mr-Quin:
## Summary
Move agent-friction knowledge out of per-developer memory into the committed docs a fresh agent already reads, so it's pre-warned instead of rediscovering the rake:
- **AGENTS.md** Gotchas: build output dir depends on the command (`vite build` -> `dev/chrome`, `pnpm run build` -> `build/`); and `pnpm <script>` can resolve into a *sibling* worktree (pin with absolute cd + the local binary).
- **browser-verify** skill: install the dir you actually just built, and confirm freshness over trusting the build banner (a stale `build/` silently runs old code).
- **e2e/AGENTS.md**: the boot-seeding race that flakes update-flow specs, and the seed-real-version + bump-the-catalog pattern that makes them deterministic.

All three are real friction hit this session.

## Test plan
- Verified: lint N/A (markdown/skill docs aren't linted) · tests N/A · e2e N/A · manual: 0 em dashes, 0 absolute machine paths (the AGENTS portability rule), well-formed, technically accurate (build-dir behavior confirmed live).
- Docs/skill only; nothing for a reviewer to run.


Mr-Quin:
Task linked: [DA-620 Capture recurring agent gotchas in committed docs (AGENTS + skills)](https://app.clickup.com/t/90131020449/DA-620)

PR #483, opened 2026-06-08
(chore) babysit-pr: screen for settled re-flags before evaluating [DA-621]

Mr-Quin:
## Summary
- `babysit-pr` 2b now screens for **settled re-flags** before evaluating: if a new open thread re-raises a suggestion already declined + resolved on this PR, resolve it with a one-line back-reference instead of a full accept/decline pass. On a multi-push PR that re-evaluation was most of the loop's wasted cost.
- Deliberately did **not** change the `ai-rereview` CI trigger: gemini re-reads the whole diff on every push regardless, so the script can't stop the re-flagging without not-re-requesting at all, and the treadmill is on code PRs (a docs-only skip wouldn't touch it). The real, low-risk fix is making the agent's handling cheap.

## Test plan
- Verified: lint N/A (skill markdown) · tests N/A · e2e N/A · manual: 0 em dashes, well-formed, no absolute paths; complements the existing one-liner in `reviewing-ai-feedback`.
- Skill-doc only; nothing for a reviewer to run.

### DIFF
--- a/.claude/skills/babysit-pr/SKILL.md
+++ b/.claude/skills/babysit-pr/SKILL.md
@@ -42,7 +42,9 @@ One GraphQL call returns: reviews, pending reviewers, open threads, reactions, c
 
 ### 2b. Classify each new comment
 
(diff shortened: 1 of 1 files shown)


Mr-Quin:
Task linked: [DA-621 Tame the ai-rereview treadmill](https://app.clickup.com/t/90131020449/DA-621)

PR #484, opened 2026-06-08
(extension) bump dango to 0.6.0: tolerate failed danmaku segments [DA-619]

Mr-Quin:
## Summary
- Bump `@mr-quin/dango` and `@mr-quin/dango-manifests` from `^0.5.0` to `^0.6.0`.
- dango 0.6.0 makes concurrent `forEach` steps fail fast (one failed iteration dooms the step) and adds an opt-in `RunOptions.continueOnError`. A danmaku fetch fans out across segments, so add `DANMAKU_RUN_OPTIONS` (`continueOnError: true`) and pass it ONLY to `runDanmaku`. One missing segment no longer drops the whole overlay.
- search / season / episodes / parseUrl and the login probe keep `MANIFEST_RUN_OPTIONS` (no `continueOnError`), so they still fail hard.

## Test plan
- [ ] `pnpm --filter @mr-quin/danmaku-anywhere test` (unit: `ManifestProviderService.test.ts` pins `runDanmaku` to `DANMAKU_RUN_OPTIONS`, the other pipelines to `MANIFEST_RUN_OPTIONS`, and asserts the base options never gain `continueOnError`).
- [ ] `pnpm --filter @mr-quin/danmaku-anywhere test:e2e -- specs/sources/tencent.spec.ts`

## Notes
- e2e turned out to be feasible and was added: `specs/sources/tencent.spec.ts` gains a partial-failure case that 404s one of two danmaku segments and asserts the surviving segment's comments still render. Verified non-theatre by inverting the implementation back to `MANIFEST_RUN_OPTIONS`, which makes the new e2e (and the unit call-site assertion) fail while the happy path stays green.
- New POM helper `SeasonDetailsPage.expectCommentCountToBe` scopes to the count caption rather than the flattened episode-button text (which merges the episode number with the count).

### DIFF
--- a/packages/danmaku-anywhere/e2e/pom/SeasonDetailsPage.ts
+++ b/packages/danmaku-anywhere/e2e/pom/SeasonDetailsPage.ts
@@ -1,10 +1,12 @@
 import { expect, type Locator, type Page } from '@playwright/test'
 
+// Matches the comment-count label in either zh ("条弹幕") or en ("comments").
+const COMMENT_LABEL = '条弹幕|comments?'
+
 const SELECTORS = {
   episodeForProvider: (provider: string) =>
     `[data-testid^="episode-list-item-${provider}-"]`,
-  // Matches either zh ("X条弹幕") or en ("X comments") locales.
-  COMMENT_COUNT_RE: /\d+\s*(条弹幕|comments?)/i,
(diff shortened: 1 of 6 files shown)


Mr-Quin:
Task linked: [DA-619 Bump dango to 0.6.0: opt into continueOnError for runDanmaku (fail-fast forEach)](https://app.clickup.com/t/90131020449/DA-619)

PR #485, opened 2026-06-08
(extension) e2e: doctrine fixes, assertion hardening, proxy env defaults [DA-611]

Mr-Quin:
## Summary
- **migration-smoke render gate:** replaced the swallowed `waitForFunction(...).catch(() => undefined)` with a real `expect(page.locator('[role="treeitem"]').first()).toBeVisible()`, so a post-migration render error fails the test instead of passing silently. Verified on both the legacy seeded probe and the post-swap probe.
- **UI-anchored `install.spec.ts`:** it asserted only `da.*` state; now also opens the popup providers list and asserts the three seeded providers render under their localized names.
- **Sanctioned `occlusion-cache-scope.spec.ts`:** documented in its header as a browser-capability probe (OPFS scope has no DOM signal), per `e2e/AGENTS.md`.
- **Mount assertion hardening:** the copy-pasted raw `[role="alert"]` import-success assertion in `import-detach` / `import-from-url` now uses the existing `ImportResultDialog.expectSuccess()` (testid-scoped, locale-proof); the loose `not.toHaveText('0')` comment-count checks in `preload-next` / `refresh-danmaku` now pin the UI count to the exact stored `commentCount`.
- **E2E proxy env defaults at build time:** a fresh clone with no `.env` produced a broken e2e build (no DNR proxy rule + degraded catalog seed). Inject placeholder `VITE_PROXY_URL`/`VITE_PROXY_ORIGIN` in `vite.config.ts` gated on the e2e env, and drop the CI step that wrote a synthetic `.env`. An exported env var still wins.

## Test plan
- [ ] `pnpm --filter @mr-quin/danmaku-anywhere test:e2e` (full suite; CI runs this)
- Ran locally and passing: the 3 doctrine specs (install, occlusion-cache-scope, migration-smoke), the 4 mount specs (import-detach, import-from-url, preload-next, refresh-danmaku), and the proxy/DDP/auth-sensitive source specs (dandanplay, dandanplay-custom, dandanplay-local-origin, login-probe, catalog-import).
- Verified the e2e build inlines the placeholder proxy host and a developer `.env.local` staging host does not leak into the e2e build.

## Notes
- The proxy env-defaults change makes local and CI e2e builds reproducible from a clean clone; it deterministically uses placeholder hosts for e2e (the mocks are path-based, so the host value is irrelevant to the tests).
- Scope grew from the original two doctrine fixes to fold in the highest-value bounded audit follow-ups. The `da`/`seed` fixture refactor (touches ~37 specs) and feature-coverage gaps remain tracked separately.

### DIFF
--- a/.github/workflows/quality-e2e.yml
+++ b/.github/workflows/quality-e2e.yml
@@ -90,16 +90,6 @@ jobs:
           restore-keys: |
             prior-releases-${{ runner.os }}-
 
-      # .env is gitignored, so write a synthetic one for the build. The
-      # hosts are intercepted by Playwright routes; nothing reaches them.
-      - name: Write .env for build
-        working-directory: packages/danmaku-anywhere
-        run: |
-          cat > .env <<'EOF'
-          VITE_PROXY_URL=https://api.danmaku.test
-          VITE_PROXY_ORIGIN=https://danmaku.test
-          EOF
-
       - name: Build extension (e2e — dev API on, open shadow root)
         working-directory: packages/danmaku-anywhere
         env:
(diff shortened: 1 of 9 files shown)


Mr-Quin:
Task linked: [DA-611 e2e suite quality and maintainability improvements](https://app.clickup.com/t/90131020449/DA-611)

PR #512, opened 2026-06-30
(app) preload host libwayland in the release-manager AppImage [DA-657]

Mr-Quin:
## Summary
- Fixes the release-manager AppImage coming up as a **blank white window** on newer Linux hosts (reproduced on Fedora 44, Wayland, Intel Arc / Mesa 26).
- Root cause: the AppImage bundles ubuntu-22.04's `libwayland-*`. On a much newer host the stale `libwayland-client`/`libwayland-egl` can't negotiate EGL with the host compositor/Mesa, so the WebKit web process aborts at `Could not create default EGL display: EGL_BAD_PARAMETER`. No `WEBKIT_*` flag helps because the failure is below WebKit, in Wayland/EGL init.
- Fix: when launched from the AppImage, `main.rs` preloads the host's `libwayland` and re-execs once (LD_PRELOAD only applies at process start). Guarded to run only inside the AppImage (`APPDIR` set) and only once (`RM_HOST_WAYLAND`). Falls through to normal startup if no host libs are found.

## Scope / not fixed here
- This makes the AppImage **launch on its own** (no manual `LD_PRELOAD`). It does **not** fix performance: the bundled ubuntu webkit still renders in software on new GPUs, so it can feel sluggish. The real fix is to use the **host** webkit (ship rpm/deb, or a non-bundling AppImage); deferred past v1 and noted in `AGENTS.md`.
- Code-only change: no CI or signing changes, so the updater signature flow is untouched.

## Test plan
- `cargo check -p release-manager` clean.
- Confirmed manually that `LD_PRELOAD=/usr/lib64/libwayland-client.so.0 <appimage>` renders (this PR automates that).
- [ ] After merge, re-cut `release-manager-v0.1.0` and confirm the AppImage launches with no manual env vars on Fedora/Wayland.

### DIFF
--- a/app/release-manager/AGENTS.md
+++ b/app/release-manager/AGENTS.md
@@ -44,6 +44,13 @@ unauthenticated requests suffice.
   `NO_STRIP=1` for `tauri build`. linuxdeploy bundles an older `strip` that
   cannot parse that section and fails otherwise. The CI runner (ubuntu-22.04)
   does not need this.
+- Host webkit vs bundled: the AppImage bundles the ubuntu-22.04 webkit and
+  libwayland. On a much newer host the bundled libwayland can't negotiate EGL
+  with the host compositor/Mesa, so the window fails to start. `main.rs`
+  preloads the host libwayland and re-execs once (only when launched from the
+  AppImage) to work around it. The bundled webkit still renders in software on
+  newer GPUs, so it can feel sluggish; the real fix is to use the host webkit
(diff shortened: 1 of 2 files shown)


Mr-Quin:
Task linked: [DA-657 Standalone release-manager desktop app (Tauri, self-updating)](https://app.clickup.com/t/90131020449/DA-657)

PR #513, opened 2026-07-23
(app) build a Windows NSIS installer in release CI [DA-657]

Mr-Quin:
## Summary
- Adds a `windows-latest` job to the release-manager workflow that builds a **Windows NSIS installer** and uploads it to the **same draft release** as the Linux AppImage.
- **Installer only** (per decision): a small config override (`src-tauri/tauri.windows.conf.json`) sets `bundle.createUpdaterArtifacts: false`, so the Windows job produces no updater artifacts, **needs no signing key**, and Windows ships **without in-app self-update** (users re-download to update).
- The Windows job `needs: build` so the Linux job creates the draft first, avoiding a concurrent create-release race.
- No app code changes: the core already compiles on Windows (unix-only bits, `config.json` mode `0600` and the libwayland trampoline, are `cfg`-gated). Windows uses **WebView2**, so the Linux bundled-webkit/EGL problems do not apply.

## Why installer-only
Windows self-update would require merging Windows into the shared `latest.json` (both platforms in one release's manifest). Skipped for now to keep CI simple; can be added later.

## Known limitation
The in-app updater UI still exists on the Windows build. Auto-check on mount swallows errors, but the manual "Check for updates" button will error there (the feed's `latest.json` has no `windows-x86_64` entry). Hiding the updater UI on Windows is a small follow-up if wanted.

## Test plan
- Workflow YAML + override JSON validated locally.
- [ ] On the next `release-manager-v*` tag, confirm the draft gets both the Linux `.AppImage` and the Windows `.exe` (NSIS) installer.
- [ ] Reviewer with Windows: run the installer, confirm the app launches (WebView2) and the Stable/Installed/Browse UI works. Not verifiable in this Linux environment.

### DIFF
--- a/.github/workflows/release-manager.yml
+++ b/.github/workflows/release-manager.yml
@@ -8,11 +8,14 @@ name: Release Manager App
 # Flow:
 #   1. Push a release-manager-v<version> tag (or run manually) -> the build job compiles,
 #      signs, and creates a DRAFT GitHub release with the AppImage, its signature, and a
-#      latest.json updater manifest. Nothing is published automatically.
+#      latest.json updater manifest. A second job adds a Windows NSIS installer to the
+#      same draft (installer only, no self-update, so no updater artifacts or signing).
+#      Nothing is published automatically.
 #   2. A human reviews and PUBLISHES that draft release.
 #   3. Publishing fires the promote job, which copies latest.json onto the floating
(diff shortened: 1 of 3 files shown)


Mr-Quin:
Task linked: [DA-657 Standalone release-manager desktop app (Tauri, self-updating)](https://app.clickup.com/t/90131020449/DA-657)

PR #515, opened 2026-07-24
(extension) bundle dango manifest catalog as offline fallback [DA-661]

Mr-Quin:
Bundles the dango manifest catalog into the build as an **offline fallback** for manifest loading. Closes DA-661.

## Problem
A fresh install with the catalog proxy unreachable gets zero sources: `onInstalled → syncCatalog` fails, the manifest store stays empty, `seedDefaultProviders` refuses to seed, and recovery waits for the 12h alarm or a manual refresh.

## What it does
- **Bundle** all `@mr-quin/dango-manifests` JSONs into the build via a Vite eager glob (`bundledCatalog.ts`); the index is derived from each manifest's own fields (same shape the network catalog produces). Adds ~104KB uncompressed to the background bundle.
- **Fallback:** `ManifestRegistry.update()` seeds missing manifests from the bundle **only when the network index is unreachable**, tagged with a new `kind: 'bundled'`. Remote stays primary; a reachable catalog always wins (fresh).
- **Auto-upgrade:** the first successful sync replaces any still-listed `'bundled'` entry with the remote copy (clearing the tag) — no manual "Update". `getPendingUpdates` skips bundled entries so they never surface as a manual update.

## Tests
- Unit (`ManifestRegistry.test.ts`): offline seeds bundled + tagged; reachable index doesn't touch the bundle; auto-upgrade replaces + clears the tag; bundled entries excluded from pending updates.
- e2e (`providers/offline-fallback.spec.ts`): first-boot offline → built-ins load from bundle (`checked never`); then proxy recovers with a bumped manifest → refresh auto-upgrades to `preinstalled` + new version, asserted against storage ground truth, no manual click.

## Verify
`build:packages` + extension build OK (manifests confirmed inlined). Full Vitest 735/735. e2e providers/sources/baseline/lifecycle 45/45. biome clean. `tsc` adds no new errors (the 4 pre-existing Zod v4 / react-hook-form resolver errors on master are unrelated).

### DIFF
--- a/packages/danmaku-anywhere/e2e/network/catalog.ts
+++ b/packages/danmaku-anywhere/e2e/network/catalog.ts
@@ -91,3 +91,15 @@ export function mockCatalog(
     },
   }
 }
+
+// Simulates a fully unreachable proxy: both the index and file endpoints
+// fail. Used to exercise the offline bundled-catalog fallback in
+// ManifestRegistry, which only kicks in once the index request fails.
+export function offlineCatalog(): NetworkMock {
+  return {
+    pattern: /\/manifest(\/file)?(\?|$)/,
+    respond: async (route: Route) => {
+      await route.fulfill({ status: 500, body: 'unavailable' })
+    },
+  }
+}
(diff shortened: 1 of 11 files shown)


Mr-Quin:
Task linked: [DA-661 Bundle manifests into the build as offline catalog fallback](https://app.clickup.com/t/90131020449/DA-661)

PR #519, opened 2026-07-26
(proxy) shorten manifest cache TTL to 15 minutes [DA-665]

Mr-Quin:
## Summary
- Drop the cache TTL on both dango manifest routes from 1 hour to 15 minutes.
- Tests pin the emitted `Cache-Control` so the TTL cannot drift silently.

## Why

A manifest fix published upstream kept being served stale long after it landed, which leaves the extension showing a pending update that reapplying does not clear. A shorter TTL bounds that window. The underlying index/file skew is structural and tracked separately.

## Test plan
- Verified: lint pass for the touched files · tests `pnpm --filter proxy test` -> 66 passed (13 files) · e2e skipped
- [ ] `pnpm --filter proxy test`

## Notes
- Needs an accompanying CDN configuration change to take effect, and proxy deploys are manual. Details in the task.
- e2e skipped: no extension-facing behavior change, and the e2e catalog specs serve their own fixtures rather than hitting the proxy.

### DIFF
--- a/backend/proxy/src/routes/api/manifest/router.test.ts
+++ b/backend/proxy/src/routes/api/manifest/router.test.ts
@@ -35,6 +35,7 @@ describe('Manifest API', () => {
 
     expect(response.status).toBe(200)
     expect(fetchSpy).toHaveBeenCalledWith(`${dangoBaseUrl}/catalog.json`)
+    expect(response.headers.get('Cache-Control')).toBe('max-age=900')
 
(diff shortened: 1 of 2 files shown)


Mr-Quin:
Task linked: [DA-665 proxy: shorten manifest cache TTL to 15 minutes](https://app.clickup.com/t/90131020449/DA-665)

PR #520, opened 2026-07-27
(chore) upgrade @hookform/resolvers so type-check survives either zod hoist [DA-666]

Mr-Quin:
## Summary
- Bump `@hookform/resolvers` from `^5.2.2` to `^5.4.3` in the extension package, so `zodResolver` type-checks regardless of which zod v4 minor pnpm hoists.

## Context

Two zod v4 minors coexist in the tree. `4.4.3` is what the extension and its runtime dependencies (`ai-sdk`, `better-auth`, `@hono/zod-openapi`) use. `4.3.6` arrives through the web app's Angular CLI toolchain, via `@modelcontextprotocol/sdk`.

zod bakes its own version into its declarations: `zod/v4/core/versions.d.ts` ships `minor: 3` in 4.3.6 and `minor: 4` in 4.4.3. `@hookform/resolvers@5.2.2` defaults its generic to the fully branded `z4.$ZodType<Output, Input>`, so a schema built by one copy fails the overload against the other with `Type '4' is not assignable to type '3'`.

Which copy `@hookform/resolvers` resolves its types from is not stable across installs from the same lockfile. Measured, all with an unmodified lockfile:

| Checkout | resolvers | `zod/v4/core` resolves to | `tsc --noEmit` |
| --- | --- | --- | --- |
| long-lived local checkout | 5.2.2 | 4.3.6 and 4.4.3 | 4 errors |
| long-lived local checkout | 5.4.3 | 4.4.3 only | clean |
| freshly bootstrapped worktree | 5.2.2 | 4.4.3 only | clean |

So this is not a broken master. CI installs fresh, lands on the good hoist, and PR Quality is green on master and on branches cut from it. What it is: a contributor whose checkout happens to hoist 4.3.6 gets four `TS2769` errors in `SignInForm.tsx:26`, `SignUpForm.tsx:34`, `AiProviderForm.tsx:96` and `IntegrationEditor.tsx:54` that CI will not reproduce, and `pnpm lint` runs `tsc` first so lint fails with it. That is an unpleasant thing to debug, and it is luck that CI has been on the right side of it.

## Why 5.4.3

Upstream fixed exactly this in resolvers PR 852, "zodResolver() type overload fails with Zod v4.3.x due to branded version mismatch", which replaces the branded generic default with a minimal structural interface that carries no version brand. That shipped in `5.4.2`, so `5.4.3` is the smallest bump that actually contains the fix rather than merely happening to work.

Not going further to `5.5.x` is a deliberate choice to keep the bump minimal, not a claim that `5.4.3` is more mature. Both lines are recent: `5.4.1` through `5.4.3` and `5.5.0` were all published on 2026-07-25. An earlier revision of this PR described `5.4.3` as a May release. That was wrong, `5.4.0` is from May and the patch releases are not.

Reviewed the release notes across `5.2.2` to `5.4.3`: nothing breaking. `5.4.1` made several validator libraries optional peer dependencies, which is what the large `peerDependenciesMeta` block in the lockfile diff comes from. Type-only, no runtime effect. `zodResolver` is used in exactly the four files listed above and nowhere else in the repo.

## Test plan
- Verified: lint pass (exit 0) · type-check pass (exit 0) · tests `pnpm test` -> 957 passed across 7 packages (integration-policy 10, danmaku-converter 57, danmaku-engine 46, danmaku-provider 36, proxy 66, web-scraper 7, danmaku-anywhere 735) · e2e skipped, no runtime behavior changes and no spec touched
- [ ] Reviewer sanity check: `pnpm exec tsc --noEmit` in `packages/danmaku-anywhere` is clean on this branch

## Notes
- No source files changed. The diff is `package.json` plus the lockfile.
- The caret was checked against how CI actually installs. `pr-quality.yml` runs a bare `pnpm install`, and pnpm leaves a lockfile-pinned version alone when it already satisfies the range, so `^5.4.3` does not float to `5.5.7` on CI.
- Deliberately not addressed here: deduplicating zod itself. A root `pnpm.overrides` entry collapsing 4.3.6 into 4.4.3 would remove the ambiguity at its source rather than making one consumer tolerant of it, but it forces the Angular CLI toolchain onto a zod minor it was not resolved against. That is a larger call and belongs in its own change.

### DIFF
--- a/packages/danmaku-anywhere/package.json
+++ b/packages/danmaku-anywhere/package.json
@@ -58,7 +58,7 @@
     "@fontsource-variable/noto-sans-sc": "^5.2.10",
     "@fontsource-variable/noto-sans-tc": "^5.2.10",
     "@fontsource-variable/plus-jakarta-sans": "^5.2.8",
-    "@hookform/resolvers": "^5.2.2",
+    "@hookform/resolvers": "^5.4.3",
     "@mediapipe/tasks-vision": "0.10.18",
     "@mr-quin/dango": "^0.8.1",
     "@mr-quin/dango-manifests": "^0.8.1",
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -355,8 +355,8 @@ importers:
         specifier: ^5.2.8
         version: 5.2.8
       '@hookform/resolvers':
-        specifier: ^5.2.2
-        version: 5.2.2(react-hook-form@7.76.0(react@19.2.0))
+        specifier: ^5.4.3
+        version: 5.4.3(@standard-schema/spec@1.1.0)(ajv@8.20.0)(react-hook-form@7.76.0(react@19.2.0))(zod@4.4.3)
       '@mediapipe/tasks-vision':
         specifier: 0.10.18
         version: 0.10.18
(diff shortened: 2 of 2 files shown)


Mr-Quin:
Task linked: [DA-666 Pin zod to a single version to unbreak type-check on master](https://app.clickup.com/t/90131020449/DA-666)

Mr-Quin:
## Sweep verdict: RELEVANT-AND-CORRECT

Verified against master `b457aa69`. No changes pushed.

**Still needed.** Master's `packages/danmaku-anywhere/package.json:61` is still `^5.2.2` and its lockfile still
carries both `zod@4.3.6` and `zod@4.4.3`. On a long-lived local checkout of master, `pnpm exec tsc --noEmit` in
`packages/danmaku-anywhere` fails right now with exactly 4 `TS2769`, one per `zodResolver` call site
(`IntegrationEditor.tsx:54`, `AiProviderForm.tsx:96`, `SignInForm.tsx:26`, `SignUpForm.tsx:34`), ending in
`Type '4' is not assignable to type '3'`. The branch worktree type-checks clean.

**The fix is the version, not a lucky hoist.** The branch worktree hoists 4.4.3 by default, so I forced the
failing condition into it: copied a real `zod@4.3.6` into `<resolvers 5.4.3 realpath>/node_modules/zod`.
`--traceResolution` confirms the resolvers package then resolved `zod/v4/core` with
`Package ID 'zod/v4/core/index.d.cts@4.3.6'` while the app files stayed on 4.4.3, and `tsc --noEmit` still
reported 0 errors. A standalone repro outside the repo shows the same setup failing on 5.2.2 and passing on
5.4.3. Bisecting the published tarballs, `5.4.0` and `5.4.1` still declare `T extends z4.$ZodType`; `5.4.2` is
the first with the unbranded local `Zod4Type`, so the body's account of where the fix landed is right.

**Inference is not weakened by the structural interface.** With 5.4.3, `form.watch('email')` is still `string`,
`setValue` still rejects an unknown field name (verified with a consumed `@ts-expect-error`).

**Lockfile is faithful.** `pnpm install --lockfile-only` on the branch prints `Already up to date` with an empty
diff. The lockfile change introduces exactly one new tarball; the two optional peers it now links
(`@standard-schema/spec@1.1.0`, `ajv@8.20.0`) were already in master's tree.

**One correction to the body.** "No runtime behavior changes" is not quite true in general: 5.4.3 changes the
`invalid_union` path, picking the branch with the fewest issues and re-pathing nested union issues under the
union's own key (5.2.2 leaked them as top-level siblings). It is an upstream bug fix and it does not reach this
repo, because the only union in the schema sources is the `z.discriminatedUnion('provider', ...)` at
`aiProviderConfig/schema.ts:54`, and both resolver versions produce byte-identical errors for it. Nothing to
change; noting it so the claim is not cited later.

All checks green at the head: `quality (24.x)`, `e2e (24.x)`, i18n, both Workers Builds, PR title.

PR #521, opened 2026-07-27
(proxy) parse trusted origins as a list [DA-671]

Mr-Quin:
## Summary

- `BETTER_AUTH_TRUSTED_ORIGINS` is a comma-separated env var, but `trustedOrigins: [env.BETTER_AUTH_TRUSTED_ORIGINS]` wrapped it as a single array element, so the literal joined string could never match a real `Origin` header. Only the `baseURL` origin (which better-auth trusts implicitly) was ever actually trusted.
- Added `parseTrustedOrigins` in `backend/proxy/src/auth/trustedOrigins.ts`: splits on commas, trims each entry, drops empty segments.
- `config.ts` now passes `parseTrustedOrigins(env.BETTER_AUTH_TRUSTED_ORIGINS)` to better-auth instead of the single-element array.

## Behavior change (please read)

This is not parse-only in effect: the corrected parse newly trusts origins in staging and production that are not functionally trusted today.

- Production `BETTER_AUTH_TRUSTED_ORIGINS` is `https://danmaku.weeblify.app,https://api.danmaku.weeblify.app`, and `baseURL` is `https://api.danmaku.weeblify.app`. Today only `api.danmaku.weeblify.app` is trusted (as the implicit baseURL origin). After this change, `https://danmaku.weeblify.app` becomes trusted too.
- Staging trusts `https://danmaku-staging.weeblify.app` and `https://danmaku.weeblify.app` in addition to its baseURL, for the same reason.
- Dev is unaffected: `advanced.disableOriginCheck` is already `true` there, so the origin check is skipped regardless.

These are exactly the origins already listed in the env var, i.e. the ones the config was always intended to trust, so this is the fix doing its job rather than an unrelated widening. Flagging it here per the acceptance criteria since it is a real change in which origins better-auth will accept for cookie-based requests.

## CORS credentials follow-up

Checked whether `hono/cors` needs `credentials: true` for cookie-based auth from the web app. It currently does not set it (`backend/proxy/src/index.ts`), and the only existing better-auth client (`packages/danmaku-anywhere/src/common/auth/createAuthClient.ts`) is bearer-token only, so this has not mattered yet. Once a cookie-based client (the web app) is added, it will need `credentials: true` on the CORS middleware plus `credentials: 'include'` on its fetch calls, or the browser will refuse to expose the response. Not changed in this PR as instructed; leaving as a follow-up.

## Test plan

- Verified: lint clean on changed files (pre-existing `noVar` errors in `public/pages/reset-password.html` are unrelated to this change) · tests `backend/proxy (affected) -> 15 files / 72 passed` · e2e skipped (backend change, no spec applies)
- `backend/proxy/src/auth/trustedOrigins.test.ts`: unit coverage of the parse helper itself: multi-origin split, whitespace trimming, empty-segment dropping, empty input.
- `backend/proxy/src/auth/config.test.ts`: integration coverage that exercises better-auth's actual origin-check path, not just the string helper. Builds a real `betterAuth` instance through `drizzleAdapter`/D1 and sends a cookie-bearing `POST /auth/sign-out` from `https://b.example`: with `trustedOrigins: [rawValue]` (the old bug) it gets rejected with a 403 `Invalid origin`; with `trustedOrigins: parseTrustedOrigins(rawValue)` it's accepted. Confirmed by mutation: reverting `parseTrustedOrigins` to `(v) => [v]` turns the "accepts" case back into a 403 failure.

### DIFF
--- a/backend/proxy/src/auth/config.test.ts
+++ b/backend/proxy/src/auth/config.test.ts
@@ -1,0 +1,54 @@
+import { env } from 'cloudflare:test'
+import { betterAuth } from 'better-auth'
+import { drizzleAdapter } from 'better-auth/adapters/drizzle'
+import { describe, expect, it } from 'vitest'
+import { parseTrustedOrigins } from '@/auth/trustedOrigins'
+import { getOrCreateDb } from '@/db'
+
+const BASE_URL = 'https://api.test.example'
+const RAW_TRUSTED_ORIGINS = 'https://a.example,https://b.example'
+
+function createTestAuth(trustedOrigins: string[]) {
+  return betterAuth({
+    baseURL: BASE_URL,
+    basePath: '/auth',
+    secret: 'test-secret',
+    database: drizzleAdapter(getOrCreateDb(env.DB), { provider: 'sqlite' }),
+    trustedOrigins,
+    advanced: {
+      disableOriginCheck: false,
+    },
+  })
+}
+
+function signOutRequest(origin: string) {
+  return new Request(`${BASE_URL}/auth/sign-out`, {
+    method: 'POST',
+    headers: {
+      Origin: origin,
+      Cookie: 'better-auth.session_token=whatever',
+    },
+  })
+}
+
+describe('trustedOrigins wired into betterAuth', () => {
+  it('rejects a cookie-bearing request from a listed origin when the list is passed unparsed', async () => {
+    const auth = createTestAuth([RAW_TRUSTED_ORIGINS])
+
+    const response = await auth.handler(signOutRequest('https://b.example'))
+    const body = await response.json()
+
+    expect(response.status).toBe(403)
+    expect(body.message).toBe('Invalid origin')
+  })
+
+  it('accepts a cookie-bearing request from a listed origin when the list is parsed', async () => {
+    const auth = createTestAuth(parseTrustedOrigins(RAW_TRUSTED_ORIGINS))
+
+    const response = await auth.handler(signOutRequest('https://b.example'))
+    const body = await response.json()
+
+    expect(response.status).toBe(200)
+    expect(body.success).toBe(true)
+  })
+})
--- a/backend/proxy/src/auth/config.ts
+++ b/backend/proxy/src/auth/config.ts
@@ -2,6 +2,7 @@ import { betterAuth } from 'better-auth'
 import { drizzleAdapter } from 'better-auth/adapters/drizzle'
 import { admin, bearer, jwt, openAPI } from 'better-auth/plugins'
 import { ac, adminRole, moderatorRole, userRole } from '@/auth/permissions'
+import { parseTrustedOrigins } from '@/auth/trustedOrigins'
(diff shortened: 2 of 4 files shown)


Mr-Quin:
Task linked: [DA-671 Split BETTER_AUTH_TRUSTED_ORIGINS into a list instead of one joined string](https://app.clickup.com/t/90131020449/DA-671)

Mr-Quin:
## agent-verified

### Runtime evidence

This is a Cloudflare Worker route with no browser surface, so a browser session is the wrong instrument. The correct equivalent is exercising the real handler, and that is what the added coverage does: it builds actual `betterAuth` instances with `advanced.disableOriginCheck: false` and sends a cookie-bearing request through `auth.handler()`, so better-auth's own origin-check middleware runs rather than a stub.

### Revert-fails

Reverting `parseTrustedOrigins` to `(value) => [value]` and rerunning `src/auth`: **5 of 6 fail** (baseline 6 passed).

The load-bearing one is `accepts a cookie-bearing request from a listed origin when the list is parsed`, which is the integration test against the real handler, not a string assertion. That is the difference between proving the parse produces the right array and proving better-auth actually accepts a request it previously rejected with 403.

The single survivor is the empty-value case, which returns the same result under both implementations and therefore protects nothing on its own. It is harmless alongside the four discriminating cases.

### Notes

The behavior change is real and intended: origins configured in the env var were never functionally trusted before, because the whole comma-joined string was passed as one array element and better-auth compares origins exactly. Only the baseURL origin was trusted. After this, the configured frontend origins are trusted as intended.

Mr-Quin:

- Still needed: `origin/master:backend/proxy/src/auth/config.ts:50` is still `trustedOrigins: [env.BETTER_AUTH_TRUSTED_ORIGINS]`, and the var is a joined list in every environment (`wrangler.json:16`, `:89`, `:167`).
- No undefined-input risk from the new `.split`: `BETTER_AUTH_TRUSTED_ORIGINS` is a wrangler `vars` entry, typed non-optional in `worker-configuration.d.ts`.
- Backend suite on this head: `Test Files 15 passed (15)`, `Tests 72 passed (72)`.

One gap worth knowing before merge, not a blocker. The new tests do not pin the line that was actually broken. `config.test.ts` builds its own `betterAuth` instance instead of going through the shipped `getOrCreateAuth`, so reverting `config.ts:51` back to `trustedOrigins: [env.BETTER_AUTH_TRUSTED_ORIGINS]` still leaves the whole suite green (`15 passed`, `72 passed`). I measured that rather than assuming it.

I did not close the gap, because it is not cheap. Calling the real `getOrCreateAuth(env)` under `@cloudflare/vitest-pool-workers` dies with `Error: Secret "BETTER_AUTH_SECRET_STG" not found`: the secret comes from a `secrets_store_secrets` binding, and miniflare only accepts `{store_id, secret_name}` there, with no way to inject a literal test value. Closing it means seeding a local secrets store or adding a seam in `config.ts`, plus handling the module-level `auth` memo. That is a follow-up task, not a change to land on a green PR.

PR #522, opened 2026-07-27
(proxy) validate the kazumi rules file path [DA-670]

Mr-Quin:
## Summary

- `GET /kazumi/rules/file` validated its `file` query param with only `z.string().min(1)`, then interpolated it unescaped into an upstream `raw.githubusercontent.com` URL, so a value like `../../../etc/passwd` traversed outside the intended `KazumiRules` directory.
- Anchored the schema to `z.string().regex(/^[\w.-]+\.json$/)`, mirroring the existing pattern used by the dango manifest file route (`backend/proxy/src/routes/api/manifest/router.ts`). The disallowed `/` blocks traversal, encoded slashes, leading slashes, and absolute URLs alike, since the validated value can never contain a path separator.
- Confirmed against the live upstream directory listing that every one of the 82 real `.json` rule files still matches the new pattern, so legitimate rule fetching is unaffected.

## Test plan

- Verified: lint clean on changed files (pre-existing unrelated warnings/errors in `public/pages/reset-password.html` and `drizzle.config.ts` predate this change) · tests `pnpm --filter proxy test -> 13 files / 67 tests passed` · e2e skipped (backend change, no spec applies)
- Added a rejection test covering a naive `../` traversal, a `%2f`-encoded slash bypass, a leading slash, and an absolute URL, all asserted to return 400 with `fetch` never called. Verified this test fails (500, not 400) when the regex is reverted to `min(1)`, confirming it actually protects the fix.

## Notes

`tsgo --noEmit` currently fails to run against `backend/proxy` on master too (a pre-existing `tsconfig.json` incompatibility with the installed `tsgo` version, unrelated to this change), so type-check for this package could not be run through the normal script. The root `pnpm type-check` (which does not cover this package, since it lacks a `type-check` script) passes cleanly.

### DIFF
--- a/backend/proxy/src/routes/api/kazumi/rules.test.ts
+++ b/backend/proxy/src/routes/api/kazumi/rules.test.ts
@@ -56,27 +56,25 @@ describe('Kazumi Rules API', () => {
     expect(content.success).toBe(false)
   })
 
-  it('rejects path traversal and encoded bypass attempts (GET /file)', async () => {
+  it.each([
+    '../../../etc/passwd',
+    '..%2f..%2f..%2fetc%2fpasswd',
+    '..%252f..%252fetc%252fpasswd',
+    'a.json%00../../etc/passwd',
+    '..%5c..%5cx.json',
+    '/etc/passwd',
+    'https://evil.com/malicious.json',
+  ])('rejects %s as the file parameter (GET /file)', async (file) => {
     const fetchSpy = vi
       .spyOn(globalThis, 'fetch')
       .mockRejectedValue(new Error('fetch should not be called'))
 
-    const maliciousFiles = [
-      '../../../etc/passwd',
-      '..%2f..%2f..%2fetc%2fpasswd',
-      '/etc/passwd',
-      'https://evil.com/malicious.json',
-    ]
+    const request = new Request(
+      createTestUrl(`/kazumi/rules/file?file=${file}`)
+    )
+    const response = await makeUnitTestRequest(request)
 
-    for (const file of maliciousFiles) {
-      const request = new Request(
-        createTestUrl(`/kazumi/rules/file?file=${file}`)
-      )
-      const response = await makeUnitTestRequest(request)
-
-      expect(response.status).toBe(400)
-    }
-
+    expect(response.status).toBe(400)
     expect(fetchSpy).not.toHaveBeenCalled()
   })
 })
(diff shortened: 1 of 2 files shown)


Mr-Quin:
Task linked: [DA-670 Validate the kazumi rules file path against a pinned pattern](https://app.clickup.com/t/90131020449/DA-670)

Mr-Quin:
## agent-verified

### Runtime evidence

Worker route, no browser surface, so the equivalent is exercising the route itself. The tests drive the real handler and assert that `fetch` is never called on a rejected input, not merely that a 4xx came back. That distinction matters here: a 4xx returned after the upstream request had already been issued would still mean the Worker had been used as a fetch proxy.

### Revert-fails

Reverting `file: z.string().regex(filePattern)` to `z.string().min(1)` and rerunning `src/routes/api/kazumi`: **7 of 10 fail** (baseline 10 passed). Every reject case fails, and because they run through `it.each` each failure names its own payload:

```
rejects ../../../etc/passwd
rejects ..%2f..%2f..%2fetc%2fpasswd
rejects ..%252f..%252fetc%252fpasswd     (double-encoded)
rejects a.json%00../../etc/passwd        (null byte)
rejects ..%5c..%5cx.json                 (backslash)
rejects /etc/passwd
rejects https://evil.com/malicious.json
```

The 3 survivors are the accept-path tests, which is correct: they should pass under both implementations.

### Notes

The opposite failure mode was also checked during review, and it is the more dangerous one: a pattern that rejects legitimate rule files would break the feature entirely. The live upstream directory was enumerated and every `.json` file there matches `^[\w.-]+\.json$`, so nothing legitimate is refused.

Mr-Quin:

- Still needed: `origin/master:backend/proxy/src/routes/api/kazumi/rules.ts` is unchanged (`file: z.string().min(1)`); last commit in that directory is `110f9abb`.
- The test is a real guard, not decoration. Reverting the one production line to `min(1)` and re-running `rules.test.ts` gives `Tests 7 failed | 3 passed (10)` with `AssertionError: expected 500 to be 400`, all seven payloads.
- Full backend suite on this head: `Test Files 13 passed (13)`, `Tests 73 passed (73)`. Biome on both changed files: `Checked 2 files in 19ms. No fixes applied.`
- No legitimate request is rejected: all 18 names in the live `index.json` and all 84 `.json` files in the upstream directory match the new pattern.

One correction to the PR summary, since it affects how severe this reads. `../../../etc/passwd` is not reachable. `fetch()` normalizes dot segments in the URL parser, so the value resolves to `https://raw.githubusercontent.com/etc/passwd`, and the host cannot be swapped either (`?file=https://evil.com/x` stays under the raw host as a path segment). The real exposure is an arbitrary-path fetch proxy against `raw.githubusercontent.com`, which is exactly what the Notes section of DA-670 already says. The fix is still right on those terms; only the summary overstates it.

PR #523, opened 2026-07-27
(extension) verify message origin on the app postMessage bridge [DA-667]

Mr-Quin:
## Summary

- Gate the app bridge `message` listener on `event.source === window` and on `event.origin` matching an allowed app origin, before any payload is trusted. Previously any page holding a handle to an app window could post a command into the switch, including `setRequestHeaders`, which installs a session DNR rule with a caller-controlled `urlFilter` and a forced `Access-Control-Allow-Origin: *`.
- Move the app URL list out of `manifest.ts` into `src/common/appUrls.ts` so the manifest `content_scripts.matches` and the runtime origin check read the same list.
- Trust `http://localhost:4200` in dev builds only. `getAppUrls(isDev)` is the single gate: `manifest.ts` passes the build-context flag, `appOrigin.ts` passes `import.meta.env.DEV`. A shipped extension therefore neither injects the bridge on localhost nor accepts it as an origin, so a locally running dev server that renders untrusted content cannot drive the bridge on a real user's machine.
- Add `src/content/app/appOrigin.ts`, which normalizes through `new URL(...).origin` and then matches with the existing `matchUrl` (URLPattern), not string comparison.
- Add `src/content/app/index.test.ts`, which drives the real registered window listener rather than the predicate in isolation.

## Test plan

- Verified: lint pass (tsc 0 errors, biome 16 pre-existing warnings, 0 new, no fixes applied) · tests `pnpm exec vitest run` in packages/danmaku-anywhere -> 78 files / 749 tests passed, of which the new `src/content/app/index.test.ts` is 14 · e2e skipped, no e2e spec touched and no e2e coverage exists for the app bridge
- Mutation checked the new tests against the source, not just against themselves:
  - delete the `isAllowedAppOrigin` check from `index.ts` -> 9 of 14 fail
  - delete the `event.source !== window` check from `index.ts` -> 1 of 14 fails
  - make `getAppUrls` return the dev list unconditionally -> 1 of 14 fails
- Reject cases go through the listener with a correct `source: 'app'` payload, so they only pass because the guard runs: `https://evil.com`, `https://danmaku.weeblify.app.evil.com`, `https://evil.danmaku.weeblify.app`, `https://evil-quinfish.workers.dev`, `https://quinfish.workers.dev.evil.com`, `http://danmaku.weeblify.app`, `http://localhost:4201`, `null`, plus an allowed origin posted from an iframe window, plus `http://localhost:4200` with `DEV` stubbed false.
- Accept cases cover prod, a staging subdomain, and local dev in a dev build, plus prod still accepted with `DEV` stubbed false (so the production-build test is not passing by breaking everything).
- Build agreement checked on real output. Production `vite build`: `content_scripts[0].matches` is `["https://danmaku.weeblify.app/*", "https://*.quinfish.workers.dev/*"]` and `grep -r localhost:4200 build/` finds nothing, so the pattern is dead-code eliminated from the bundle too. `NODE_ENV=development vite build`: the manifest and the emitted content script both carry `http://localhost:4200/*`.

## Notes

- The origin check reuses `matchUrl` (URLPattern) so a wildcard host pattern never degrades into a substring test. One intentional difference from Chrome match-pattern semantics: `https://*.quinfish.workers.dev/*` matches the bare host `quinfish.workers.dev` for content script injection but not for this origin check, since URLPattern requires the leading label. Staging always runs on a subdomain, so the effect is that the bridge is slightly stricter than injection, never looser.
- The app posts with `window.postMessage(payload, window.location.origin)` to its own window, so `event.source === window` and `event.origin` is the app origin. Existing flows (kazumiSearch, kazumiGetChapters, extractMedia, episodeGetAll, danmakuGet, setRequestHeaders) are unchanged.
- Dev-workflow consequence of the localhost gate: driving a local `app/web` dev server now needs a dev-mode extension (`pnpm dev`, or `pnpm dev:browser` without `--build`). A production build run locally will no longer talk to `localhost:4200`. This matches how `externally_connectable` already gates `http://localhost:4321` on `isDev`.

### Known adjacent gap, not closed by this PR

There is a second, wider route to the same background RPC surface, raised separately. `manifest.ts` `externally_connectable.matches` is `*://danmaku.weeblify.app/*`, whose scheme wildcard includes plaintext http, and `RpcManager.ts` binds the full background RPC server (including `setHeaders`) to `chrome.runtime.onMessageExternal`. That list is also duplicated in `src/common/constants.ts`, the same drift this PR removed for the app URLs. Do not read this PR as having sealed the whole surface.

Also still out of scope as agreed: tightening the DNR rule built in `packages/web-scraper/src/messaging/messaging.ts` (initiator/tab restriction, expiry, payload schema validation). That path still has no runtime validation, it is just no longer reachable from a foreign page.

### DIFF
--- a/packages/danmaku-anywhere/manifest.ts
+++ b/packages/danmaku-anywhere/manifest.ts
@@ -1,6 +1,6 @@
 import { defineManifest } from '@crxjs/vite-plugin'
 import { getBuildContext } from './scripts/getBuildContext'
-import { APP_URLS } from './src/common/appUrls'
+import { getAppUrls } from './src/common/appUrls'
 
 const { isDev, browser, appVersion } = getBuildContext()
 
@@ -43,7 +43,7 @@ export const manifest = defineManifest({
   },
   content_scripts: [
     {
-      matches: APP_URLS,
+      matches: getAppUrls(isDev),
       js: ['src/content/app/index.ts'],
       run_at: 'document_start',
     },
--- a/packages/danmaku-anywhere/manifest.ts
+++ b/packages/danmaku-anywhere/manifest.ts
@@ -1,5 +1,6 @@
 import { defineManifest } from '@crxjs/vite-plugin'
 import { getBuildContext } from './scripts/getBuildContext'
+import { APP_URLS } from './src/common/appUrls'
 
 const { isDev, browser, appVersion } = getBuildContext()
 
@@ -24,12 +25,6 @@ if (isChrome) {
 if (isDev) {
   permissions.push('declarativeNetRequestFeedback')
 }
-
-const APP_URLS = [
-  'https://danmaku.weeblify.app/*', // prod
-  'https://*.quinfish.workers.dev/*', // staging
-  'http://localhost:4200/*', // local dev
-]
 
 export const manifest = defineManifest({
   manifest_version: 3,
--- a/packages/danmaku-anywhere/src/common/appUrls.ts
+++ b/packages/danmaku-anywhere/src/common/appUrls.ts
@@ -1,0 +1,7 @@
+// Consumed by manifest.ts at build time, so this module must stay free of
+// browser and import.meta.env dependencies.
+export const APP_URLS = [
+  'https://danmaku.weeblify.app/*', // prod
+  'https://*.quinfish.workers.dev/*', // staging
+  'http://localhost:4200/*', // local dev
+]
--- a/packages/danmaku-anywhere/src/common/appUrls.ts
+++ b/packages/danmaku-anywhere/src/common/appUrls.ts
@@ -1,7 +1,17 @@
-// Consumed by manifest.ts at build time, so this module must stay free of
-// browser and import.meta.env dependencies.
-export const APP_URLS = [
+// Consumed by manifest.ts in the Node build context, so this module must stay
+// free of browser and import.meta.env dependencies.
+const APP_URLS = [
   'https://danmaku.weeblify.app/*', // prod
   'https://*.quinfish.workers.dev/*', // staging
+]
+
+const DEV_APP_URLS = [
   'http://localhost:4200/*', // local dev
 ]
+
+export function getAppUrls(isDev: boolean): string[] {
+  if (isDev) {
+    return [...APP_URLS, ...DEV_APP_URLS]
+  }
+  return APP_URLS
+}
--- a/packages/danmaku-anywhere/src/content/app/appOrigin.ts
+++ b/packages/danmaku-anywhere/src/content/app/appOrigin.ts
@@ -1,8 +1,13 @@
-import { APP_URLS } from '@/common/appUrls'
+import { getAppUrls } from '@/common/appUrls'
 import { matchUrl } from '@/common/utils/matchUrl'
+import { tryCatchSync } from '@/common/utils/tryCatch'
 
 export function isAllowedAppOrigin(origin: string): boolean {
-  return APP_URLS.some((pattern) => {
-    return matchUrl(origin, pattern)
+  const [url, err] = tryCatchSync(() => new URL(origin))
+  if (err) {
+    return false
+  }
+  return getAppUrls(import.meta.env.DEV).some((pattern) => {
+    return matchUrl(url.origin, pattern)
   })
 }
--- a/packages/danmaku-anywhere/src/content/app/appOrigin.ts
+++ b/packages/danmaku-anywhere/src/content/app/appOrigin.ts
@@ -1,0 +1,8 @@
+import { APP_URLS } from '@/common/appUrls'
+import { matchUrl } from '@/common/utils/matchUrl'
+
+export function isAllowedAppOrigin(origin: string): boolean {
+  return APP_URLS.some((pattern) => {
+    return matchUrl(origin, pattern)
+  })
+}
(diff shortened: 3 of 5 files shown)


Mr-Quin:
Task linked: [DA-667 Verify message origin on the app postMessage bridge](https://app.clickup.com/t/90131020449/DA-667)

Mr-Quin:
## agent-verified

### Runtime evidence, real browser

Verified in a real Chromium with the unpacked extension loaded and its service worker running, not in jsdom. The MCP browser available here lacks the Extensions CDP domain, so this was driven through Playwright, the same mechanism the e2e suite uses.

The actual attack shape was staged rather than approximated: a page served on `http://localhost:4201` embedded inside the app page on `http://localhost:4200`, posting a forged `{source: 'app'}` request (`episodeGetAll`) into the app window. That gives the content script a message whose `event.source` is a foreign window and whose `event.origin` is a non-allowlisted origin, which is exactly the reachable path this PR closes.

Two builds, identical test:

| Build | Legit post from 4200 | Forged post from 4201 |
| --- | --- | --- |
| this branch | responded, `success: true` | **no response** |
| guards stripped, rebuilt | responded, `success: true` | **responded, `success: true`** |

So the vulnerability reproduces on the unguarded build and is closed on this one. The legitimate app path keeps working in both, which was the real risk of an over-strict origin check: a fix that silently broke the web app would have passed every unit test in the repo.

### Revert-fails

Removing the `isAllowedAppOrigin(event.origin)` call from `src/content/app/index.ts` fails **9 of 17** tests in `src/content/app` (baseline 17 passed).

### Notes

Scope is the `window.message` entry point only. `externally_connectable` remains `*://` and still binds the full RPC surface to `onMessageExternal`, tracked separately. This PR should not be read as sealing that route.

Mr-Quin:
**Sweep verdict: RELEVANT-AND-CORRECT.** Nothing changed or pushed.

I reproduced the vulnerability against a real production-mode build of master in a real Chrome, then ran the same attack against this branch.

**Master is still exploitable, and it is the shipped build that is exposed.** `origin/master` `b457aa69` `src/content/app/index.ts:60-64` still starts with `if (event.data?.source !== DA_EXT_SOURCE_APP) return` and nothing else, and `manifest.ts:28-32` still hardcodes `http://localhost:4200/*` with no `isDev` gate. Building master with plain `vite build` (`isDev: false`, `daEnv: prod`) yields:

```
master prod matches: ["https://danmaku.weeblify.app/*","https://*.quinfish.workers.dev/*","http://localhost:4200/*"]
```

Loading that build in Playwright chromium, with an attacker page on `http://127.0.0.1:4200` that opens `http://localhost:4200/` and posts commands into the window handle:

```
### MASTER PROD BUILD
SAME-WINDOW PROBE: {"ok":true,"success":true,"source":"content-script"}
VICTIM SAW RESPONSES TO FOREIGN POSTS: [{"id":"foreign-episodeGetAll","type":"response","source":"content-script","success":true},{"id":"foreign-setRequestHeaders","type":"response","source":"content-script","success":false}]
```

`episodeGetAll` ran to a successful response for a foreign origin, and `setRequestHeaders` reached the handler too (it came back failed for my payload shape, not ignored). So the DNR path in the task description is reachable from web content today.

Methodology note if anyone re-runs this: the bridge replies with `window.postMessage` on the app window, not to `event.source`, so an attacker-side listener sees nothing even on a successful attack. My first probe read "no response" on master and looked clean; the observable has to be read inside the victim page.

**Same attack against this branch.** The dev build is the harder test, since that is the build where `localhost:4200` is still an allowed origin and only the runtime guard can stop it:

```
### PR 523 DEV BUILD (localhost allowed)
SAME-WINDOW PROBE: {"ok":true,"success":true,"source":"content-script"}
VICTIM SAW RESPONSES TO FOREIGN POSTS: []
```

**No regression on the real app origins.** `event.source === window` is the one line jsdom cannot honestly verify (no isolated world), so I checked it in Chrome with the branch's **production** build, serving the probe page at the real origins through request interception:

```
SAME-WINDOW PROBE @ https://danmaku.weeblify.app: {"ok":true,"success":true,"source":"content-script"}
SAME-WINDOW PROBE @ https://danmaku-anywhere.quinfish.workers.dev: {"ok":true,"success":true,"source":"content-script"}
```

**Build agreement, measured both ways.** Default `vite build`: manifest matches are prod plus staging only, and `grep -rn "localhost:4200"` over the entire output finds nothing, so the pattern is constant-folded out of the content script as well. `NODE_ENV=development vite build`: localhost is present in the manifest *and* in the emitted `assets/index.ts-*.js`. `manifest.ts` gets `isDev` from `getBuildContext` (`NODE_ENV === 'development'`) and `appOrigin.ts` gets `import.meta.env.DEV`, which Vite derives from the same `NODE_ENV`, so the two sides cannot drift in the loosening direction.

**The tests protect the guard.** `pnpm exec vitest run src/content/app` -> 2 files / 17 tests passed. Mutating the shipped source one change at a time reproduces your numbers exactly: dropping the `isAllowedAppOrigin` check gives `9 failed | 5 passed (14)`, dropping the `event.source !== window` check gives `1 failed | 13 passed (14)`, making `getAppUrls` return the dev list unconditionally gives `1 failed | 13 passed (14)`. All mutations reverted, worktree clean.

All five DA-667 acceptance criteria are met. The criterion asks for the origins to come from "the existing manifest constant"; moving that constant into `src/common/appUrls.ts` and importing it from `manifest.ts` satisfies the one-list intent.

The adjacent gap you flagged is real and still open on master: `manifest.ts:65-71` `externally_connectable.matches` is `*://danmaku.weeblify.app/*` (scheme wildcard includes http), duplicated at `src/common/constants.ts:4`. Out of scope here, worth its own task.

PR #524, opened 2026-07-27
(extension) seed default providers when the catalog is unreachable [DA-668]

Mr-Quin:
## Summary

- Seed the default provider configs on both sync outcomes. An offline first install registered the bundled manifests but created zero configs, because `syncCatalog()` returned early on an unreachable index and never reached `seedDefaultProviders()`. The user landed on an empty sources list with no explanation.
- Fall back to the bundled catalog whenever the index yields no usable manifests, not just when it is unreachable. `ManifestRegistry.update()` checked `loadIndex()` for null, but a parseable index with an empty `manifests` array (or one whose entries are all dropped by the supported-api-version filter) returns `[]`, which is truthy, so the bundle was skipped.
- Never seed over configs that already exist. The seeded flag and the configs both live in `chrome.storage.sync`, which on a signed-in profile can still be arriving at boot, so an offline first run could read the flag as unset and push the three defaults out to every other device. The seed now bails and locks the flag when any config is present. That also makes the non-atomic write safe: a worker killed between `options.set()` and `markSeeded()` finds configs on the next sync and locks instead of re-seeding.
- Stop locking the flag unconditionally on an update install. A user whose first run was offline on the old code has no configs and no flag, and would have been locked at zero sources forever. Flag unset plus zero configs is not a user who deleted their configs (that user's flag is set), so that pair stays unlocked and the sync heals it.
- Report a catalog that answers but lists nothing this build can run with its own message instead of the fetch-failure one. `update()` now returns `'synced' | 'unreachable' | 'empty'` and `refreshCatalog()` picks the matching error.
- Concurrent syncs (install, alarm, popup refresh) share one in-flight seed, so the asynchronous flag read cannot be raced into a double write within a process.

## Test plan

- Verified: lint `tsc --noEmit` 0 errors + `biome ci` 0 errors (16 pre-existing warnings in untouched files) · tests `vitest run` in packages/danmaku-anywhere -> 744 passed / 77 files, including 9 new cases (2 in `ManifestRegistry.test.ts` for the empty index and the all-filtered index; 7 in `ProviderService.test.ts` for the offline seed, the locked flag, two concurrent syncs, the non-empty-config guard, healing a never-seeded install on update, not re-seeding a deleted-config install on update, and the distinct empty-catalog error) · e2e `e2e/specs/providers/offline-fallback.spec.ts` (2) and `e2e/specs/sources/update-recovery.spec.ts` (3, it asserts the refresh error toast this PR touches) -> 5 passed, rebuilt with `VITE_DA_ENV=e2e` first; rest of the suite left to CI per the affected-only rule.
- Reverted the two source files and re-ran: the 2 new registry cases, the offline-seed case, the concurrency case, and the rewritten first e2e test all fail. Reverted the non-empty guard and the update-install change on their own: `locks the flag instead of writing over configs that already exist` and `seeds an install whose first run never reached the seed when it updates` both fail.
- One new case passes either way by construction and is kept as a regression guard, not as proof: `does not seed on an unreachable catalog once the flag is locked` (on the old code that path never seeds at all).

## Notes

- `offline-fallback.spec.ts` previously documented the bug (it asserted the built-ins showed Import buttons, with a comment noting the seed never ran). It now asserts the fixed behavior: the three built-ins are installed, "Installed (3)" renders, no Import button is offered for them, and the configs are enabled. The second test lost its `markSeeded()` workaround and both tests route the login probes the seeded sources fire on render, since `applyProfile` cannot be used here (it clears the storage this spec is about).
- The non-empty-config guard narrows the sync-overwrite window but cannot close it completely: configs arriving between the guard's read and `options.set()` would still be overwritten. Closing that fully needs a merge-on-arrival strategy for `chrome.storage.sync`, which is beyond this fix.
- Healing a never-seeded install re-seeds one rare pre-flag case: a user who deleted every config on a build that predates the seeded flag and then upgrades straight to this version. Any install that passed through a flag-aware version already had the flag locked on that upgrade, so the exposure is limited to that skip-upgrade path, and it is preferred over leaving genuine offline-install victims at zero sources.

### DIFF
--- a/packages/danmaku-anywhere/e2e/pom/ProvidersPage.ts
+++ b/packages/danmaku-anywhere/e2e/pom/ProvidersPage.ts
@@ -67,6 +67,12 @@ export class ProvidersPage {
     return this.page.getByText(/^(Update failed|更新失败)$/)
   }
 
+  installedHeading(count: number): Locator {
+    return this.page.getByText(
+      new RegExp(`^(Installed|已安装) \\(${count}\\)$`)
+    )
+  }
+
   checkedNeverLabel(): Locator {
     return this.page.getByText(/not checked yet|尚未检查/)
   }
--- a/packages/danmaku-anywhere/e2e/specs/providers/offline-fallback.spec.ts
+++ b/packages/danmaku-anywhere/e2e/specs/providers/offline-fallback.spec.ts
@@ -1,24 +1,32 @@
+import type { BrowserContext } from '@playwright/test'
 import {
   manifestVersion,
   mockCatalog,
   offlineCatalog,
 } from '../../network/catalog'
+import { mockLoginProbes } from '../../network/loginProbes'
 import { Popup } from '../../pom/Popup'
 import { expect, test } from '../../setup/fixtures'
 
 /**
  * First-boot offline fallback: the manifest catalog proxy is unreachable
  * (index and file endpoints both fail) on a fresh profile with an empty
- * manifest store. ManifestRegistry seeds its bundled dango manifests instead
- * of leaving the registry empty, so the built-in sources still appear in the
- * Providers page catalog with no successful network round trip, and the
- * catalog is reported as never having been checked. Once the proxy recovers,
- * the next reachable sync auto-replaces a bundle-seeded manifest with the
- * catalog copy with no manual apply, since the tag on it (not surfaced as a
- * pending update) exists only to mark it as not user-chosen.
+ * manifest store. ManifestRegistry seeds its bundled dango manifests and the
+ * boot sync still seeds the default provider configs from them, so the
+ * built-in sources land installed and usable with no successful network round
+ * trip, and the catalog is reported as never having been checked. Once the
+ * proxy recovers, the next reachable sync auto-replaces a bundle-seeded
+ * manifest with the catalog copy with no manual apply, and does not re-seed
+ * the configs on top of the offline ones.
  */
 
 const BUMPED = '9.9.9'
+
+const BUILT_IN_NAMES = [
+  /DanDanPlay|弹弹play/,
+  /Bilibili|B站/,
+  /Tencent Video|腾讯视频/,
+]
 
 test.use({
   catalogMock: offlineCatalog(),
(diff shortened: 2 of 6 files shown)


Mr-Quin:
Task linked: [DA-668 Seed default providers when the manifest catalog is unreachable](https://app.clickup.com/t/90131020449/DA-668)

Mr-Quin:
Sweep verdict: **FIXED-AND-PUSHED**. The content was already relevant and correct; the red check was not caused by this diff.

**The red check.** `Workers Builds: danmaku-anywhere-web` failed on the old head `13831693` (build `736d23bf-1e1d-4c72-bf56-f7a2244ad927`, `2026-07-27T08:25:19Z`). This diff touches only `packages/danmaku-anywhere/**`, and `app/web` has no dependency on the extension package. I rebuilt the web app at that exact head locally (`pnpm -r -F "./packages/**" build`, then `pnpm exec ng build` in `app/web`): both exit 0, "Application bundle generation complete", only the pre-existing `libass-wasm is not ESM` warning that master emits too. Every other open PR pushed on `2026-08-05` has that same check passing, and the same push that failed it passed `danmaku-anywhere-ui-preview` (`688ccc5a`), which does the same repo-root install.

I rebased the branch onto `b457aa69` (both commits replayed, no conflicts, no content change, `13831693` -> `c1d18f5c`) and force-pushed with `--force-with-lease` to retrigger. The rebuild is green:

```
Workers Builds: danmaku-anywhere-web	pass	0	.../builds/97895677-6622-43b8-9fbf-5c192f52f542
```

**Relevance.** Both halves of DA-668 are still open on master. `ProviderService.ts:420-424` still early-returns on `if (!fetched)` before `seedDefaultProviders()` at line 448, and `ManifestRegistry.ts:174-180` still falls back to the bundle only when `loadIndex()` returns null, which `[]` is not. Nothing merged since the branch point touches either file.

**Correctness.** No defects found. The review follow-up commit is real hardening, not churn: bailing out and locking when `getAll()` is non-empty closes a genuine bug, since the flag and the configs share `chrome.storage.sync` and a signed-in profile can read the flag as unset while the user's real configs are still arriving, and it also makes the non-atomic set-then-`markSeeded` pair safe across a worker kill. Dropping the unconditional lock on an `update` install is right for the same reason: a user whose first run was offline on the shipped code has no flag and no configs, and locking them would freeze them at zero sources forever. The shared-promise seed guard is sound, since both callers await the same promise so a rejection still reaches each.

Local verification on the rebased head: `tsgo --noEmit` in `packages/danmaku-anywhere` exit 0, `tsgo --noEmit` in `app/web` exit 0, `biome ci` on the extension exit 0 (16 pre-existing `noExplicitAny` warnings in untouched files), `vitest run src` 77 files / 744 tests passing. Nothing in the tree was edited; `c1d18f5c` is your tree, replayed.

PR #525, opened 2026-07-27
(chore) fix the standalone build flag type mismatch [DA-673]

Mr-Quin:
## Summary
- `vite.standalone.config.ts` and `vite.config.ts` defined `VITE_STANDALONE` as `JSON.stringify(true/false)` (a boolean), while `isStandalone.ts` compared it against the string `'true'`, so the explicit flag never matched and detection silently fell through to a `typeof chrome === 'undefined'` heuristic.
- Changed both configs to encode the flag as a string (`JSON.stringify('true')` / `JSON.stringify('false')`) and kept the string comparison in `isStandalone.ts`, so the two sides agree.
- Changed `VITE_STANDALONE` in `vite-env.d.ts` from an optional `'true' | 'false'` to a non-optional one. Note this is a documentation-only improvement, not compile-time enforcement: `ImportMetaEnv` has no link to a vite config's `define` block, so a config that forgets to set the flag is still `undefined` at runtime, not a type error. Confirmed by deleting the define from `vite.standalone.config.ts` and running `tsc --noEmit`, which still returned 0 errors. I looked at adding a runtime assert in `isStandalone.ts` to give this real enforcement, but decided against it: every current consumer of the flag lives under `packages/danmaku-anywhere/src` and goes through one of the two vite configs, both of which set it correctly, so there is no live path that needs guarding today, and an assert for a case that can't currently happen is exactly the kind of defensive code this fix should avoid. If a future third build config is added and forgets the flag, the type in `vite-env.d.ts` won't catch it.
- Removed the `typeof chrome === 'undefined' || !chrome.runtime` fallback in `isStandalone.ts` now that the explicit flag is reliable.
- Added a gotcha to `packages/danmaku-anywhere/AGENTS.md` documenting that a boolean `define` inlines to a real literal under `vite build` and `vite dev`, but surfaces as a truthy string specifically under vitest, so `import.meta.env` flags should be string-encoded, for the next person who touches a build-time flag in this package.

## Test plan
- Verified: lint tsc+biome -> pass · tests full package suite (735 tests) -> pass · e2e skipped (build config change, no runtime spec applies)
- Proved the fix with the actual define pipeline rather than just reading the diff: ran a one-off assertion of `IS_STANDALONE_RUNTIME === true` through `vitest run --config vite.config.ts` (fails, resolves to `false`, as expected for the extension build) and through `vitest run --config vite.standalone.config.ts` (passes, resolves to `true`). Both runs share the same jsdom/mocked-chrome environment, so the differing, correct results prove the outcome is now driven by the build-time flag and not by any runtime heuristic.
- Also confirmed both real production builds still succeed: `pnpm run build` (extension, emits to `build/`) and `vite build --config vite.standalone.config.ts` (standalone, emits to `build-standalone/`).
- Before the fix, running the full `vitest run` suite deterministically failed 11 tests in `ManifestStore.test.ts`: the flag resolved to a truthy non-empty string under vitest specifically (vitest stringifies `import.meta.env` custom defines regardless of the value's original shape, so a boolean-shaped define surfaces as the plain string `"false"`, which is truthy), which flipped `getStorageArea` onto the standalone `localStorage`-backed store instead of the mocked `chrome.storage.local`, leaking state across tests. This is now fixed as a side effect of encoding the flag as a proper string. Separately confirmed via the actual dev server (`vite --config vite.config.ts`) that a boolean-shaped define serves as a real, unquoted boolean there too, matching production build behavior; only vitest differs.

## Notes
The heuristic (`typeof chrome === 'undefined' || !chrome.runtime`) is gone rather than kept as a fallback. It only ever produced the right answer by coincidence: the extension build always has `chrome.runtime` available and the standalone build never does, so it happened to line up with what the broken flag comparison should have returned. Once the flag actually works, the heuristic becomes a second, independent path to the same conclusion with no case in this codebase where it's reachable with a different answer than the flag (every place that imports `isStandalone.ts` goes through one of the two vite configs, both of which now set the flag correctly, including under vitest). Keeping it would mean carrying dead code that could, in some future third build config that forgets to set `VITE_STANDALONE`, silently produce a different answer than the flag depending on which check runs first, exactly the failure mode this bug already caused once.


Mr-Quin:
Task linked: [DA-673 Fix VITE_STANDALONE flag type mismatch so the explicit flag works](https://app.clickup.com/t/90131020449/DA-673)

Mr-Quin:
## agent-verified

### Runtime evidence

Build-time flag with no user-facing runtime surface, so a browser session would prove nothing. The meaningful question is whether the flag now actually drives the result, given the bug was that it never fired and detection fell through to a `typeof chrome === 'undefined'` heuristic which this PR removes.

Answered directly by probing the same assertion under both configs, in the same environment, so the only variable is the config:

```
expect(IS_STANDALONE_RUNTIME).toBe(false)

vitest --config vite.config.ts             -> 1 passed
vitest --config vite.standalone.config.ts  -> 1 failed
```

The value genuinely differs by config. Since the heuristic is gone, nothing else could be producing that difference: the flag is doing the work.

### Notes

Two claims were corrected during review rather than left standing, which is worth recording because both were about the change overselling itself:

- The `'true' | 'false'` union in `vite-env.d.ts` is **not** compile-time enforced. Vite's `define` is an untyped string map with no link to `ImportMetaEnv`, proven by deleting the define and getting zero `tsc` errors. The PR body no longer claims otherwise.
- The AGENTS.md gotcha originally said a boolean define surfaces as a truthy string under both vitest and the dev server. Only vitest stringifies it. The dev server and production build both inline a real boolean. Corrected, since standing guidance that is half wrong is worse than none.

Mr-Quin:

Checked the claim against emitted bundles rather than tests, since this bug is invisible to vitest by construction.

- Master, standalone target built with `vite build --config vite.standalone.config.ts`: the heuristic survives in the output, so it is what actually decides standalone mode on the deployed site today.
  ```
  $ grep -o "typeof chrome[^;]\{0,60\}" assets/index-*.js | head -1
  typeof chrome>"u"||!chrome.runtime}const El=$1e()
  ```
  The flag comparison is constant-folded away. On this branch that function is absent from the bundle entirely; the two remaining `typeof chrome` hits are unrelated modules.
- Probe through the real define pipeline on this branch, asserting both `import.meta.env.VITE_STANDALONE` and `IS_STANDALONE_RUNTIME`: under `vite.config.ts` both fail as they should (`expected 'false' to be 'true'`), under `vite.standalone.config.ts` both pass. That is the first two acceptance criteria, via the flag path.
- `tsgo --noEmit` on the package: exit 0.
- Removing the fallback is safe here: only two vite configs exist in the package, and `manifest.ts:60` confirms content scripts run in the isolated world, so no extension-built bundle runs anywhere `chrome.runtime` is missing. The heuristic was weaker than it looked anyway, since `window.chrome` exists on ordinary Chrome pages.

The AGENTS.md gotcha is the most durable part of this diff and it reproduces: on master, a vitest probe asserting `import.meta.env.VITE_STANDALONE === 'true'` under the standalone config passes, because vitest stringifies env defines while build and dev inline the raw boolean.

One small thing: I could not reproduce the "11 tests in `ManifestStore.test.ts` failed before the fix" note. On master the comparison yields `false`, detection falls to the heuristic, which returns `false` under the mocked `chrome`, and the suite is green. Nothing depends on it, it just looks like a note from an intermediate state.

PR #526, opened 2026-07-27
(extension) assert real comment counts in source e2e specs [DA-669]

Mr-Quin:
## Summary

- `SeasonDetailsPage.expectCommentCount` matched a digit followed by a comment-count word anywhere in the episode row, so a literal "0 comments" passed. Worse, the row's flattened text concatenates the episode number with the count, so an episode numbered 1 showing "0条弹幕" reads as "10条弹幕" and satisfied even a strict positive-count regex.
- Added a `comment-count` testid to the season-details episode caption (the mount tree's episode row already carries the same testid) and scoped the count assertion to that element via `toHaveText`, so the count is asserted against the caption alone.
- Every source spec now asserts the exact count its fixture parses to: bilibili xml 3, bilibili proto 3000, dandanplay 4, dandanplay custom / loopback 2 each, tencent 3 (2 + 1 across segments), tencent partial-failure 2 (unchanged).
- The tencent happy path additionally opens the danmaku viewer from the mount tree and asserts each segment's comment text rendered ("hello", "测试" from segment 0, "world" from segment 30000), so a well-formed-but-empty result fails. New `DanmakuViewerPage` POM hangs off `Popup`.
- `ManifestProviderService.getDanmaku` validated nothing: it cast the manifest pipeline output to `CommentEntity[]`. It now validates each entry against `zCommentEntity` and **drops** the malformed ones, logging how many went and the first reason. A hard failure is kept only for output that is not usable at all: not an array, or a non-empty batch in which every entry is malformed. An empty result stays empty, since an episode with no danmaku is legitimate.
- New source spec covers the drop path end to end: a segment serving one comment with no text renders the well-formed comments and not the malformed one.

## Test plan

- Verified: lint clean (0 errors; 16 pre-existing `any` warnings, none in touched files) · type-check clean across all packages · tests `pnpm --filter '...[origin/master]' test` -> 77 files / 739 tests passed (includes 5 `ManifestProviderService.getDanmaku` cases covering drop, empty, not-an-array and all-malformed) · e2e `sources/tencent.spec.ts`, `sources/bilibili.spec.ts`, `sources/dandanplay.spec.ts`, `sources/dandanplay-custom.spec.ts`, `sources/dandanplay-local-origin.spec.ts` -> 8 passed. Only the touched specs were run; CI owns the sweep.
- Extension rebuilt with `VITE_DA_ENV=e2e` before every e2e run.

### Why drop rather than reject the batch

The first cut of this rejected the whole fetch when any comment failed the schema. That is the wrong trade for a danmaku app: losing 2999 good comments because one was malformed is a much worse outcome than silently dropping one. `8642c9b3` ("drop unknown comment modes instead of throwing") made exactly this call one release ago, changing `parseCommentEntityP` to return `null` so callers drop the bad comment rather than throwing and killing the render, with `CommentsTable` and `DanmakuRenderer` filtering accordingly. This boundary now matches that. Loudness is kept where it matters: the dropped count and the first schema issue are logged, and output that is not usable at all still throws.

### Mutation experiments

All run for real, not reasoned:

1. **Zero comments.** Emptied `barrage_list` in both tencent segment fixtures. Both tencent count assertions failed with `Expected pattern: /^3\s*(条弹幕|comments?)$/i` / `Received string: "0条弹幕"` and `/^2\s*.../ ` / `"0条弹幕"` respectively. That experiment is what surfaced the flattening false-pass: before the testid was added, a positive-count assertion passed on the zero fixture because it matched the wrapper element whose text content is `1` + `0条弹幕` = `10条弹幕`. Fixture reverted.
2. **Structurally invalid comments.** This is now a permanent spec rather than a throwaway mutation: `tencent-danmaku-segment-malformed.json` serves two comments where the second has no `content`, so the pipeline emits it without the `m` the schema requires. The spec asserts the episode lands with exactly 2 comments (1 surviving from the malformed segment + 1 from segment 30000), that "kept" and "world" render in the danmaku viewer, and that the viewer holds exactly 2 rows so the malformed one is provably absent. To confirm the spec is load-bearing rather than vacuous, the validation was temporarily bypassed (returning the raw pipeline output): the spec failed with `Expected pattern: /^2\s*(条弹幕|comments?)$/i` / `Received string: "3条弹幕"`, i.e. the malformed comment reaching the database is caught. Bypass reverted.
3. **Unmodified fixtures.** All 8 tests across the 5 touched source specs pass.

### What this validation does not cover

`zCommentEntity` checks that `p` and `m` are strings; it does not check the format of `p`. A pipeline emitting `p: "garbage"` still passes validation, persists, and counts toward the comment count while rendering nothing, which is the same class of failure this PR closes one layer up. Left as is deliberately: this is the schema the persisted episode shape and the non-manifest DanDanPlay provider already use, and tightening it is a separate change with its own compatibility question. The renderer already drops comments whose `p` it cannot parse, so the failure is degraded rendering rather than a crash.

### Does the validation change behavior for real provider responses?

Checked the real captures before choosing the strictness: `danmaku-provider/src/providers/tencent/test/barrage-60000.json` (1197 comments, every one has `content`, `time_offset` and `content_style`) and the 3000-element bilibili proto fixture (every element has text, all modes within 1-6). Nothing in either is dropped. All 13 bundled dango manifests build `p` by string concatenation and map `m` from a source field, so both are strings whenever the source field exists; JSONata only omits the key when the source field is missing outright, which is the case now dropped rather than persisted. `zCommentEntity` is the same shape the persisted episode schema already declares and that the non-manifest DanDanPlay provider already enforces.

One real behavior change worth naming: `z.object` strips unknown keys, so the DanDanPlay pipeline (whose output is the upstream comment objects passed through) no longer persists extra fields a compatible server returns alongside `cid`/`p`/`m`. Nothing reads those fields (the engine reads `p`, `m`, `cid`, `s`, and `s` is in the schema), so the effect is smaller IDB rows.


Mr-Quin:
Task linked: [DA-669 Assert real comment counts in source e2e specs and validate danmaku pipeline output](https://app.clickup.com/t/90131020449/DA-669)

Mr-Quin:
## agent-verified

### Runtime evidence

The e2e half runs in a real Chromium with the real extension loaded, and `PR E2E Quality` is green on this branch. During review the concatenation claim was also confirmed directly rather than inferred: instrumenting the happy-path spec to dump the episode row's live `textContent` produced `第1集13条弹幕`, showing the episode number sits immediately adjacent to the caption with no separator. That is why a tighter regex alone was not enough and the caption needed its own `comment-count` testid.

### Revert-fails

Reverting `ManifestProviderService.ts` and rerunning `src/background/services/providers`: **3 fail** (baseline 114 passed).

```
drops a comment missing its text and keeps the rest
throws when the pipeline emits something that is not an array
throws when every comment in a non-empty batch is malformed
```

Reverting the `BaseEpisodeListItem.tsx` testid fails all 8 count assertions across the 5 source specs, reported during review.

### Why this one mattered

The assertion it replaces passed on zero comments. It matched a digit followed by a comment-count word anywhere in the episode row, and the row's flattened text made `1` plus `0条弹幕` read as `10条弹幕`, so even a strict positive-count regex passed against an empty result. Most of the provider coverage in the suite rested on that assertion, so it was proving that a request fired and a row was written, not that a single comment parsed.

### Notes

Scope of the validation is honest about its own limits: `zCommentEntity` checks that `p` and `m` are strings, not that `p` parses. A pipeline emitting `p: "garbage"` still persists with a positive count and renders nothing. Same class of failure, one layer in, documented in the PR body rather than implied away.

Mr-Quin:
Swept as part of an open-PR review pass. **Verdict: RELEVANT-AND-CORRECT.** Nothing changed or pushed.

**Both halves of the premise still hold on master.** `SeasonDetailsPage.ts:29-31` is still
`toContainText(/\d+\s*(条弹幕|comments?)/)` with five happy-path call sites, and
`ManifestProviderService.ts:189` is still a bare cast whose output reaches `DanmakuService.upsert` with no
schema parsing anywhere on the path.

I ran the zero-comment experiment rather than reasoning about it. Serving the tencent happy path two empty
segments against an `e2e` build: master's assertion passes (`1 passed`), the branch's
`expectCommentCountToBe(episode, 3)` fails with `Received string: "0条弹幕"`.

Worth recording that master's *stricter* helper is also unsound, not just the loose one. On the same
zero-comment fixture, for an episode numbered 1:

```
episode.getByText(/^10\s*(条弹幕|comments?)$/i) → visible → 1 passed
```

That is the episode-number node `1` concatenated with the caption `0条弹幕`. So the pre-existing
`expectCommentCountToBe` could be satisfied by an episode with zero comments whenever the expected count
happens to equal `number . 0`. The `comment-count` testid plus `toHaveText` closes that, which makes the
testid the load-bearing part of the change rather than a convenience.

**The new e2e spec is load-bearing.** Bypassing the validation and rebuilding reproduces your stated failure
exactly: `Expected pattern: /^2\s*(条弹幕|comments?)$/i` / `Received string: "3条弹幕"`. Five source mutations
against `ManifestProviderService.test.ts` (bypass validation, drop the `output.length > 0` guard, never throw
on all-malformed, drop the `Array.isArray` guard, stop logging the drop) were each caught by a distinct test.

**Checks on the validation itself:**

- Read all 13 bundled manifests. Every one builds `p` by JSONata string concatenation and maps `m` from a raw
  source field; only `hanjutv` and `migu` emit `cid`, both via `$number(...)`; `dandanplay` passes upstream
  objects through. No manifest can fail the schema wholesale and trip the all-malformed throw.
- Unknown-key stripping is safe: every consumer reads only `p`/`m` (`CommentsTable`, `videoSkipParser`,
  `useExportXml`, `commentsToXml`, `parser.ts`) or the parsed form in `bindVideo`.
- Cost of the per-comment `safeParse`, measured: `n=3000 ms=2.9`, `n=20000 ms=8.6`, `n=100000 ms=11.7`.

**One thing to confirm, not a defect.** DA-669's acceptance criterion says structurally invalid comments should
"fail at the validation boundary with a clear error". A partially-invalid batch now drops and warns instead.
The reasoning in the PR body (and the `8642c9b3` precedent) is right, and a fully-malformed batch still throws.
Flagging only because it is a conscious narrowing of the written criterion.

Verified locally: `type-check` clean, `biome ci` 0 errors on the extension (16 pre-existing `any` warnings),
`ManifestProviderService.test.ts` 13 passed, `tencent.spec.ts` 3 passed against a fresh `VITE_DA_ENV=e2e`
build. CI at `e8a5b28`: `e2e (24.x)` `90 passed (3.0m)`, `quality (24.x)` success.

PR #527, opened 2026-07-27
(docs) update provider docs for the manifest source system [DA-674]

Mr-Quin:
## Summary

- Rewrote `docs/danmaku.mdx` (弹幕来源): replaced the fixed four-source list with the actual catalog model (Installed / 弹幕源目录 / 有可用更新 / 存在问题), the real default-enabled sources (DanDanPlay, Bilibili, Tencent, not DanDanPlay+MacCMS), the 12h auto-sync cadence, and pointed custom sources at the new manifest page and the correct add-menu label (VOD).
- Rewrote the "第三方" section of `docs/development/limitations.mdx`: removed the hand-written Bilibili/Tencent cookie-check subsections (no longer exist) and replaced them with the actual manifest-driven model (hosts allowlist enforced by the engine, loginProbe/cookieSet capabilities), corrected the host_permissions claim to match the extension's static broad grant rather than a per-manifest permission request.
- Updated `getting-started.mdx`'s "启用或禁用弹幕源" section to state the real default-enabled sources and the correct popup tab name (弹幕源, not 设置页面).
- Added the missing `docs/danmaku-manifest.mdx` that `ManifestEditorPage.tsx` links to: what a manifest is, where the catalog comes from (the public dango repo, synced through the proxy with a bundled offline fallback), and how to author/duplicate/test/save your own.
- Added a sidebar entry for the new page next to 弹幕来源.

Every claim was grounded directly in source: `ManifestRegistry`, `ProviderService`, `zManifest` (dango's schema.ts), the `providers/` popup components (`CatalogSection`, `InstalledList`, `NeedsAttentionCallout`, `UpdatesSection`, `ManifestEditorPage`), the zh-CN locale strings (to match exact UI labels), the backend `manifest/router.ts` (catalog source + 15 min cache), and `AlarmManager.ts` (12h refresh cadence).

## This PR alone cannot turn the Pages / Workers Builds checks green

Not a content issue, a build-config gap this PR surfaced (docs is the first docs-only PR in a while, so it's the first to trigger these checks):

- `docs/` has its own `pnpm-lock.yaml` and is not a member of the root `pnpm-workspace.yaml`, so it installs standalone. Since pnpm 10/11, the "which install scripts are allowed to run" setting (`allowBuilds`, covering `sharp`, `esbuild`, `@tailwindcss/oxide`) only lives in `pnpm-workspace.yaml`, and `docs/` had none of its own, so it fell back to the repo root's, which doesn't mention those three packages. That fails a plain local install under `docs/` today with `ERR_PNPM_IGNORED_BUILDS`.
- Fixed: committed `docs/pnpm-workspace.yaml` with `allowBuilds: true` for `@tailwindcss/oxide`, `esbuild`, and `sharp`. Verified from scratch (`rm -rf node_modules dist .astro && pnpm install --frozen-lockfile && pnpm build`, no `--ignore-workspace`): clean install, all three postinstall scripts run, all 27 pages build including `docs/danmaku-manifest/index.html`. This is needed and correct regardless of the point below, it fixes the local dev experience for anyone running `pnpm install` in `docs/`.
- Note for anyone who reproduces this locally: pnpm auto-generates an unfilled `pnpm-workspace.yaml` stub (`'@tailwindcss/oxide': set this to true or false`, etc.) as a side effect the moment it hits the ignored-builds case. That stub is not a committed file and never was, it will look "pre-existing" if you diff two independently-generated copies, but it isn't tracked and isn't on `origin/master`.

**This will not make CI green, and that's expected, not a bug in this diff.** Pulled the actual build logs via the Cloudflare API for both failing checks: the Pages project's `build_config.build_command` and the Workers Builds trigger (`trigger_uuid 967006e8-4a84-47e8-b6b8-bf7370d10d9b`, "Deploy non-production branches") both run:

```
pnpm --ignore-workspace install && pnpm build
```

`--ignore-workspace` makes pnpm ignore `pnpm-workspace.yaml` entirely, including `allowBuilds`, no matter what it contains, anywhere in the repo. Confirmed two ways: a filled-in file still fails identically under that flag, and pnpm resets the file back to the placeholder stub every time it hits this failure while that flag is set.

The actual fix is changing the Pages project's and the Workers Builds trigger's build command to drop `--ignore-workspace` (e.g. `pnpm install --frozen-lockfile && pnpm build`). That's account-level Cloudflare CI config, not in this git diff, invisible to PR review, and it would apply to every future build, not just this one, so it is out of scope for this PR to change. **Do not merge expecting these two checks to go green** until that dashboard/trigger config is updated separately; that is unrelated to whether this diff is correct.

## Test plan

- Verified: docs build clean (`astro check` 0 errors/warnings/hints, `astro build` succeeded) · internal links resolve (`/docs/danmaku-manifest` route generated and linked correctly from `docs/danmaku.mdx`, `development/limitations.mdx`, and the sidebar; the in-product link path `docs/danmaku-manifest/` now matches an existing page) · e2e skipped (docs only)
- Verified the install/build the way Cloudflare's build command would work if `--ignore-workspace` is dropped: `rm -rf node_modules dist .astro && pnpm install --frozen-lockfile && pnpm build` from `docs/`, clean pass, all 27 pages generated.
- Reproduced the actual CI failure locally: `pnpm --ignore-workspace install` in `docs/` fails with the same `ERR_PNPM_IGNORED_BUILDS` regardless of `pnpm-workspace.yaml` content.

## Notes

- Follow-ups (explicitly out of scope per the task): people occlusion, the in-video info panel, cloud backup, search history, and the collapse filter remain undocumented.
- Locale: the docs site declares an `en` locale in `astro.config.mjs`, but there are currently zero English content pages anywhere in the site (every page, not just the ones touched here, renders via Starlight's automatic "not available in this language" fallback to the zh-CN content). Nothing was left desynced by this change since there was no English content to update.
- Checked `docs/integration-policy.mdx`'s note that automatic matching is hardcoded to DanDanPlay: still accurate against `useLoadDanmaku.ts`, no change needed.
- The Pages and Workers Builds checks on this PR **will** stay red until the Cloudflare dashboard/trigger build command is updated separately to drop `--ignore-workspace` (see above); that's an account-level CI setting, not something a merge or another commit to this branch can fix.
- One claim I could not fully verify against a written spec (only inferred from `MANIFEST_DOCS_PATH` and Starlight's directory-index build output): that `https://docs.danmaku.weeblify.app/docs/danmaku-manifest/` (trailing slash, as the in-product link constructs it) is served correctly in production. The local static build produces `dist/docs/danmaku-manifest/index.html` and both the slash and non-slash variants resolve to it locally; I did not check the production host's redirect/rewrite config.


Mr-Quin:
Task linked: [DA-674 Update provider docs for the manifest-driven source system](https://app.clickup.com/t/90131020449/DA-674)

Mr-Quin:
Sweep verdict: **BLOCKED**. Content is relevant and correct; the two red checks cannot be fixed from inside this diff.

Relevance confirmed against master `b457aa69`: `ManifestEditorPage.tsx:34` still has `const MANIFEST_DOCS_PATH = 'docs/danmaku-manifest/'` with no such page in `docs/`, so the in-product link is dead today and this PR is what fixes it. The 12h cadence claim matches `AlarmManager.ts:105` (`periodInMinutes: 60 * 12`). Built this branch's `docs/` from a clean `node_modules`: `astro check && astro build` exit 0, 27 pages, `dist/docs/danmaku-manifest/index.html` present.

One correction to the root cause in the description. I pulled the actual Pages build log for the failing deployment:

```
Detected the following tools from environment: pnpm@10.11.1, nodejs@22.16.0
! Corepack is about to download https://registry.npmjs.org/pnpm/-/pnpm-11.2.2.tgz
Installing project dependencies: pnpm install
.../esbuild@0.25.4/node_modules/esbuild postinstall: Done
Done in 9.1s using pnpm v11.2.2
Executing user command: pnpm --ignore-workspace install && pnpm build
[ERR_PNPM_IGNORED_BUILDS] Ignored build scripts: @tailwindcss/oxide@4.1.7, esbuild@0.25.4, sharp@0.33.5, sharp@0.34.5
```

Two things that adds:

1. Cloudflare's own automatic install (plain `pnpm install`) succeeds and runs all four postinstalls. The `docs/pnpm-workspace.yaml` added here is what makes that step pass, so it is load-bearing, not just a local-dev nicety.
2. The build does not use Cloudflare's detected pnpm 10.11.1. Corepack overrides it with 11.2.2 because `docs/package.json` has no `packageManager` field, so it climbs to the root pin `"packageManager": "pnpm@11.2.2"` (package.json:37). That matters because pnpm 10 still honors `pnpm-workspace.yaml` settings under `--ignore-workspace` and only warns on ignored builds, while pnpm 11 ignores the file and makes it a hard error. Verified locally on a copy of this branch: pnpm 10.11.0 + `--ignore-workspace` + this `pnpm-workspace.yaml` gives exit 0 with all four postinstalls run; pnpm 11.2.2 gives exit 1 and overwrites the committed file with the placeholder stub.

Also ruled out empirically under pnpm 11 (all still exit 1): `pnpm.onlyBuiltDependencies` in package.json (pnpm warns the field is no longer read), `.npmrc` with `only-built-dependencies[]`, `.npmrc` with `strict-dep-builds=false`.

Remediation, both of which are yours to choose:

- Preferred: drop `--ignore-workspace` from the Pages build command and the `danmaku-anywhere-docs` Workers Builds trigger, for example `pnpm install --frozen-lockfile && pnpm build`. Passes locally with pnpm 11.2.2 given the file this PR adds.
- Only in-repo alternative: add `"packageManager": "pnpm@10.11.1"` to `docs/package.json` so corepack resolves the nearest package.json instead of the root pin. Verified end to end locally. Not pushed, since it freezes the docs site a pnpm major behind the rest of the repo to work around a CI command.

A `PNPM_VERSION` build variable alone would not help: the log shows corepack overriding the already-correct detected 10.11.1.

Nothing was pushed to this branch.

PR #528, opened 2026-07-27
(extension) validate backup data on restore [DA-672]

Mr-Quin:
## Summary

- Parse the backup envelope with a zod schema in `ConfigStateService.restoreState` instead of taking `backupData as BackupData`. The unchecked cast and its TODO are gone, and `restoreState` now takes `unknown`.
- Refuse an envelope whose `meta.version` is not the version this build writes, before any store is touched. `getState` now stamps the same constant it checks on read.
- Validate store payloads against schemas those stores already own. No new schemas were written.
- Restore is two phases: check every candidate store, then write only what passed. A store that fails is reported in `BackupRestoreResult.details` and never written, while the rest still restore.
- `downloadCloudBackup` returned `result.data as BackupData`, the second unchecked cast on the same read path. It now returns `unknown` and is validated at restore like any other blob.

### Coverage is 5 of 7 stores, and the gap is not harmless

Schema-covered: `extensionOptions` (`extensionOptionsSchema`), `localMatchingRule` (`namingRuleOptionsSchema`), and per-entry `mountConfig` (`mountConfigInputSchema`), `providerConfig` (`providerConfigSchema`), `integrationPolicy` (`zIntegration`).

**Not covered: `danmakuOptions` and `aiProviderConfig`.** For those two, a payload at the current version is envelope- and version-checked only, and its contents are **written to storage unvalidated**, exactly as before this PR. Junk in those two stores is persisted, not merely tolerated. `danmakuOptions` has no zod schema anywhere in the tree and writing one was out of scope. `aiProviderConfig` has one, but it cannot be used here: see the round-trip note below.

### List stores drop bad entries instead of rejecting the batch

`mountConfig`, `providerConfig` and `integrationPolicy` validate one entry at a time via `backupItemSchema`. Bad entries are dropped and counted in `details[store].droppedEntries`; the good ones still restore. On a recovery operation over the user's own backup, costing them every good config because one entry is corrupt is the worse failure, and this matches the drop-and-report policy used for comments in #526. A payload where *no* entry survives is refused instead, so a fully corrupt list cannot silently empty a store.

### Version gate

A payload older than the store's latest version is passed through untouched, because the historical schema no longer exists in the tree and the store's own migration chain is what brings it up to date. A payload newer than the store's latest version is refused, since there is no way to migrate backwards.

`ConfigStateService` still restores stores concurrently, so a runtime failure partway through can leave mixed state. Not addressed here; validation failures no longer contribute to it, since nothing is written until every payload has been checked.

## Review follow-ups

Four issues from review, all fixed in `24fac473`. The first two were reproduced before fixing and the repros are now regression tests.

1. **`OptionsService.set(data, 0)` treated version 0 as absent** (`if (!currentVersion)`), falling back to the stored version. Restoring `extensionOptions: {data: {legacyUnversioned: true}, version: 0}` wrote it stamped version 27, so zero migrations ran. `getVersion()` returns `0` for any store whose options predate versioning, so this is reachable from a real old export, and it directly falsified this PR's "older payloads are left to the migration chain" argument. Fixed to `currentVersion === undefined` in both `set` and `setInternal`. Verified: the same payload now lands at version 0 and comes out with `hotkeys`, `theme` and the rest backfilled by the migration chain.
2. **A failed migration reset a store while the restore reported success.** `upgrade()` caught the error, called `reset()`, and returned `void`. `extensionOptions: {data: null, version: 20}` turned a pre-restore `enabled: false` back into `enabled: true` and reported `{success: true}`. `upgrade()` now returns `'upgraded' | 'initialized' | 'reset'` and `restoreState` reports a `reset` as a store failure. Separately, `useBackupImport` always fired the success toast regardless of the result, which is why this looked clean in the UI; it now inspects `result.success` and names the stores that had problems.
3. **All-or-nothing list validation**, fixed as described above.
4. **`zIntegration` requires `policy.title.selector.min(1)` but `migrateV1ToV2`/`migrateV2ToV3` do not enforce it**, so a v1 or v2 era integration with an empty title selector survives migration and fails the current schema. Confirmed by running a v1 fixture through both migrations: `title.selector` comes out `[]`. With per-entry validation the blast radius is now that one integration rather than the whole store, it is counted in `droppedEntries`, and the toast surfaces it. Such a policy cannot extract a title, so it is non-functional rather than a working config being discarded. If it is the user's only integration, the store is refused and left untouched instead of being emptied.

## Test plan

- Verified: lint `biome check` exit 0, 16 pre-existing warnings, no new ones · tests `pnpm exec vitest run` in packages/danmaku-anywhere -> 78 files, 760 passed · e2e `e2e/specs/migration/` all 3 specs passed (7.3s) after a `VITE_DA_ENV=e2e` rebuild. Rest of the suite left to CI.
- `tsc --noEmit` exit 0.
- e2e scope: the migration directory covers restore end to end, including `migration-smoke` which seeds a prior v1.5.0 install through the popup restore flow and then swaps builds, so it exercises the changed toast path too.
- Three scenarios run against real store services backed by an in-memory `chrome.storage`:
  - Malformed backup (`services` a string): threw `Invalid backup format: ✖ Invalid input: expected record, received string → at services`, and `extensionOptions.enabled` was still its pre-restore value.
  - Malformed store payload inside a valid envelope: `success: false`, the affected stores reported with their zod errors and not written, the others restored.
  - Unrecognised `meta.version: 2`: threw `Unsupported backup version 2, this extension can only restore version 1 backups`, nothing written.
  - Valid current backup: `success: true`, all seven stores `{success: true, droppedEntries: 0}`.
- Reverting `ConfigStateService.ts` and `OptionsService.ts` to the previous commit fails 9 tests, including every one added for the four review items.

## Notes

**Round trip.** `backupRoundTrip.test.ts` builds the seven real store services over an in-memory `chrome.storage`, seeds each through its own public API, exports with `getState`, and restores that payload, plus the JSON stringify/parse trip an exported file actually takes.

To be precise about what it is worth: this file passes with the source changes of the first commit reverted. It is a **forward guard against over-strictness**, not coverage of the validation this PR adds. (Two of its cases do fail on revert now, but only because they assert the new `droppedEntries` field, which is not behaviour worth claiming.) It earns its place by failing when a schema is *tighter* than what the extension writes, which is precisely how it paid for itself: with `zAiProviderConfig` attached as `aiProviderConfig`'s schema, the round trip failed with `Invalid input: expected string, received object → at [1].settings.providerOptions.reasoning`. `providerOptions` is declared as a record of JSON strings with a `.transform(JSON.parse)`, so the store persists the transformed output and re-parsing it through the same schema rejects it. It is an input schema, not a validator for stored data, and the field is user-editable in `AiProviderForm`. The test still seeds a provider with `providerOptions`, so anyone who attaches that schema later gets a failing test explaining why.

**Old backups.** Checked against a real artifact rather than a constructed one. `e2e/fixtures/migration/backup.json.gz` is a genuine v1.5.0 export and is restored in a unit test: its `extensionOptions` (24 vs 27), `danmakuOptions` (8 vs 10) and `providerConfig` (1 vs 5) payloads are older than current and skip schema checks by design, while `mountConfig` (6), `integrationPolicy` (4) and `localMatchingRule` (1) are at current versions and are validated for real. All seven report success, and the `import-migrate` e2e confirms the same file still restores through the popup. No blocker found.

Two smaller compatibility points. Backups that predate the `shouldBackup` filter can carry `userAuth` or `searchHistory` entries, and entries for stores that no longer exist; both are ignored rather than treated as errors, with a test covering the unknown-store case. `meta.timestamp` is required by the envelope schema, which is safe because every version since the feature landed has written it.

**Error style.** Envelope and version failures throw, matching what `importAll` already did for unparseable JSON and what the RPC layer surfaces to the import toast. Per-store failures go into the existing `BackupRestoreResult.details`. `@danmaku-anywhere/result` is not used anywhere in the extension package, so introducing it here would have been the odd one out.


Mr-Quin:
Task linked: [DA-672 Validate backup data on restore instead of casting it](https://app.clickup.com/t/90131020449/DA-672)

Mr-Quin:
## agent-verified

### Runtime evidence

Restore runs in the background service worker, and the real-browser leg is `e2e/specs/migration/import-migrate.spec.ts`, which drives a genuine v1.5.0 backup artifact through the popup import flow in a real Chromium. `PR E2E Quality` is green on this branch. The unit layer runs the real store services over an in-memory `chrome.storage` rather than mocks of the code under test.

### Revert-fails

Two independent mutations, both run:

Reverting `ConfigStateService.ts` to master and rerunning `src/background/services/Backup`: **16 of 33 fail** (baseline 33 passed).

Reverting only `OptionsService.ts` to the pre-fix commit: **exactly 2 fail**, and they are named for the two defects review found by executing the restore path rather than reading it:

```
runs migrations on a version 0 payload instead of stamping it current
reports a store whose migration failed rather than silently resetting it
```

Those two were real bugs, not style points. `set(data, 0)` treated version `0` as absent and stamped genuinely pre-versioning payloads as current, skipping every migration, which directly contradicted this PR's own safety argument that old payloads are left to the migration chain. And a malformed older payload could trigger `reset()`, silently wiping the user's config while the import toast reported success.

### Why over-strictness was the thing to guard

For a restore path, a schema tighter than what the app itself writes is worse than the loose cast it replaces: it rejects users' own legitimate backups. `backupRoundTrip.test.ts` exists to catch exactly that, and it paid for itself immediately by rejecting `zAiProviderConfig` as a stored-data validator (it is an input schema with a `.transform(JSON.parse)`, so output does not round trip through it).

### Notes

Coverage is 5 of 7 stores, stated plainly in the PR body rather than implied to be complete. `danmakuOptions` and `aiProviderConfig` remain envelope- and version-checked only, and junk in those two is persisted rather than merely tolerated. The round-trip test is a forward guard against over-strictness, not coverage of this change, and the PR body says so.

Mr-Quin:
## Sweep verdict: RELEVANT-AND-CORRECT

Reviewed against `origin/master` at `b457aa69`. Nothing pushed to the branch.

### The problem is real and still on master

`BackupService.service.ts:38` on master still carries `// TODO: validate backup data` and
`restoreState(backupData as BackupData)`, and `downloadCloudBackup` still ends `return result.data as BackupData`.

I measured what that costs rather than arguing it. A throwaway vitest probe in the main checkout at `b457aa69`, over
the real `MountConfigService` / `ProviderConfigService` / `ExtensionOptionsService` backed by an in-memory
`chrome.storage`, restored a hand-written envelope:

```
result:                  {"success":true,"details":{"extensionOptions":{"success":true},"mountConfig":{"success":true}}}
stored mountConfig:      "i am not an array"
stored extensionOptions: {"enabled":"yes-please"}   at version 999
```

Reported to the user as a clean import. The version-999 stamp is the nastier half: the store's latest is 27, so
`migrateOptions` never runs on that install again and the poisoned object is permanent.

The byte-identical probe on this branch:

```
result: {"success":false,"details":{
  "extensionOptions":{"success":false,
    "error":"Backup holds version 999 of this store, which is newer than the supported version 27"},
  "mountConfig":{"success":false,"error":"Expected a list of entries"}}}
stored mountConfig:      []                      (unchanged default)
stored extensionOptions: {enabled:true, ...}     (unchanged defaults, version 27)
```

### Mutation testing: 8 mutations, 8 caught

Each applied to the shipped source, not to a test helper, then `pnpm exec vitest run src/background/services/Backup
src/common/options` (baseline 8 files / 106 tests green), then reverted.

| # | Mutation | Result |
|---|---|---|
| 1 | `if (meta.version !== BACKUP_FORMAT_VERSION)` to `if (false)` | 3 fail: envelope version 0 / 2 / 99 |
| 2 | `if (backupData.version > latestVersion)` to `if (false)` | 1 fail |
| 3 | `if (!service.backupSchema)` to `if (true)` | 1 fail |
| 4 | `if (service.backupItemSchema)` to `if (false)` | 4 fail across 2 files |
| 5 | `currentVersion === undefined` back to `!currentVersion` (both sites) | 1 fail: version 0 payload |
| 6 | `return 'reset'` to `return 'upgraded'` on migration failure | 1 fail |
| 7 | `if (droppedEntries > 0 && kept.length === 0)` to `if (false)` | 1 fail |
| 8 | delete `.filter((candidate) => candidate.check.ok)` | 4 fail, incl. all three refuse-and-leave-untouched cases |

Every decisive line is guarded by a test that fails without it.

### One coverage gap worth closing before merge (not a blocker)

`restores the v1.5.0 backup file shipped as an e2e fixture` is weaker than it looks. Every store version in that
fixture is below current, so `checkPayload` returns on the passthrough branch and **no schema runs on it at all**. It
proves migrations still work; it does not prove validation accepts real data.

So I wrote the test that does: restore the fixture (migrating every store to current), re-export with `getState`,
JSON round-trip, restore that. The second restore is the one that hits full validation, on data produced by the
extension's own migration chain rather than a test factory.

```
re-exported versions: extensionOptions 27, danmakuOptions 10, mountConfig 6, providerConfig 5,
                      aiProviderConfig 1, integrationPolicy 4, localMatchingRule 1
second restore:       success true, all 7 stores {success: true, droppedEntries: 0}
```

Zero drops. Validation is not stricter than the extension's own output. Suggest adding that round trip to
`backupRoundTrip.test.ts`; it is about fifteen lines on top of the fixture test already there, and it would be the
only test in the file exercising the validating branch against migrated data.

### Both stated reasons for the 5-of-7 coverage gap check out

- `src/common/options/danmakuOptions/` has `constant.ts`, `service.ts`, `useDanmakuOptions.ts` and no zod anywhere.
- `zAiProviderConfig` genuinely cannot be reused, because it is not idempotent: `settings.providerOptions` is
  `z.record(z.string(), z.string().refine(isJson).transform(JSON.parse))` (schema.ts:30-49), so its input is a JSON
  string and its output, which is what gets stored, is a parsed object. Re-parsing stored data fails on any provider
  that set `providerOptions`. Splitting that transform out is a separate change.

### Things I checked that turned out fine

- `z.unknown()` does not make the key optional in zod v4: an entry with a missing `data` key is refused
  (`safeParse` returns `success: false`), so it cannot write `undefined` into an uncovered store.
- `checkEntries` using `safeParseAsync` is load-bearing, not incidental. `mountConfigInputSchema.patterns` carries an
  async `.refine` calling `validateOrigin`; a sync parse would throw.
- Keeping the original entry (`data.filter(...)`) rather than the parsed output is the right call, since it stops
  zod's stripping and defaults from silently rewriting a user's config on the way through restore.
- `mountConfigInputSchema` is the correct schema to validate stored mount configs against, because
  `MountConfigService.create` parses with that same schema and stores the output.
- `cloudBackupDownload` becoming `RPCDef<string, unknown>` breaks nothing: its only consumer,
  `CloudBackupSection.tsx:71`, forwards the value into `importMutation.mutate(data)`, which takes `unknown`.

### Pre-existing behavior this PR does not change

A payload whose store version is below latest still passes through unvalidated and is written before migration runs;
if migration then throws, `upgrade()` resets the store and the user's pre-existing settings for it are gone. That is
master's behavior today, and the only difference here is that it is now reported as a failure instead of returning
`success: true`. Fixing it properly needs historical schemas that no longer exist in the tree, so leaving it is
right. It does mean "nothing is written until every payload has been checked" holds only for current-version
payloads.

### Verification

- `pnpm exec vitest run src/background/services/Backup src/common/options`: 8 files / 106 tests passed.
- `pnpm exec vitest run` (whole extension package): 78 files / 760 tests passed.
- `pnpm exec tsgo --noEmit -p packages/danmaku-anywhere/tsconfig.json`: exit 0.
- `pnpm exec biome check packages/danmaku-anywhere/src`: 0 errors, 14 warnings. Same command on master gives 14, and
  `diff` of the two sorted warning lists is empty, so this branch adds none. (The PR body says 16 pre-existing; the
  count is 14 on both sides.)
- CI on head `0df7dc84`: `e2e (24.x)`, `quality (24.x)`, `Validate i18n translations (24.x)`,
  `Validate PR title format`, `check-changes`, `Workers Builds: danmaku-anywhere-web`,
  `Workers Builds: danmaku-anywhere-ui-preview` all SUCCESS.

All five DA-672 acceptance criteria are met, including the TODO being deleted rather than reworded. Good to merge as
is; the re-export test is the one thing I would add first.

PR #533, opened 2026-08-19
(extension) retry occlusion cross-origin recovery instead of claiming DRM [DA-679]

### LINKED ISSUE #532: 使用ai   ——  弹幕在人物后方时报错无法加载遮罩
### 最新版本

- [x] 我使用的是最新版本

### 问题描述

disabled: video canvas is tainted (cross-origin/DRM)

Image

### 复现步骤

1.
2.
3.

### 网络请求与日志

_No response_

### 浏览器类型

Chrome

### 其他浏览器

_No response_

### 是否为移动设备

- [ ] 是

Mr-Quin:
Closes #532.

Occlusion turned itself off on a site serving a plain cross-origin mp4, reporting "this video is protected (cross-origin or DRM) and cannot be read".

## Root cause

`CrossOriginCapture.setup()` ended with `sync()`, which seeks the fresh clone to the live playhead. Seeking drops `readyState` below 2, and the readability probe treated "no decoded frame" as not clean, so a readable video was declared protected. The slower the clone loads, the larger the drift, the more reliably it misfires. Compounding it: a single failed recovery was terminal and permanent, and timeout, blocked request and genuine protection all reported the same status.

## Structural fix

The probe returned a boolean over three states (readable, tainted, no frame yet), and callers compensated with scattered `readyState` guards. One missing guard was the bug. Frame acquisition is now capability-based:

- `frameProbe.ts` returns `readable | tainted | undetermined`, so "no decoded frame" is a state the compiler forces callers to handle.
- `frameStrategy.ts` defines `FrameStrategy`, `Frame` (carrying its `mediaTime`), `FrameFailure` (`protected` vs retryable `unavailable`, with evidence) and `AcquireResult`.
- `DirectFrameStrategy` reads the live element; `CloneFrameStrategy` owns the DNR clone, runs setup in the background, and reports `pending` while a sync-initiated seek is in flight, so a pre-seek picture can no longer reach the compositor.
- `frameStrategyFactory.ts` classifies: undetermined -> pending, readable -> direct, tainted + `mediaKeys` -> protected/encrypted, tainted + same-origin -> protected/same-origin, tainted + non-http -> protected/non-http, tainted + cross-origin http -> clone.
- `frameSource.ts` is a coordinator: plan lifetime, retry backoff, staleness. Nothing else.

Same-origin media cannot taint, so a tainted same-origin source is protected content by deduction rather than assumption. Retry policy stays in the coordinator so per-failure-kind policy has one home.

Behavior changes that follow: a rejected clone setup settles as a failed attempt instead of parking the video; `reset()` disposes a recovery still in flight; encrypted media skips the clone entirely; the DNR rule drops the clone request's `Origin` header (hotlink-protected CDNs reject it) and is scoped to the requesting tab; `taint` and `unreadable` are distinct statuses with honest messages.

## Tests

66 unit tests across the probe, both strategies, the classification table, the coordinator's retry and staleness behavior, and the service's outcome handling. The clone state machine covers rejected setup, dispose mid-setup, a seeking clone, a clone with no decoded frame, and a decoded clone that is still tainted.

e2e harness routes reject the clone request on `Origin`, reject it once, or delay it; specs assert recovery through each path plus the honest give-up message when nothing can rescue the source. Full suite green (92 passed).

Not covered: that a captured bitmap matches the live playhead. `Frame` now carries `mediaTime`, which makes that assertable later; no consumer checks it yet.

## Clone lifecycle

Two ways the recovery clone could burn resources without saying anything:

- It only learned that playback stopped when the capture loop polled it, and the loop stops ticking when the video pauses or the tab hides. It now mirrors pause, play and visibility from the events directly.
- An installed clone that stopped decoding while the original played sat in `pending` forever, since `readyState` below 2 read as "mid-seek". Five seconds without a decoded frame is now a retryable failure, so it backs off, rebuilds, and reports the video unreadable if it never recovers.

## Test plan

- `pnpm lint`: pass (tsc, biome, spec theatre lint)
- unit: 771 passed across the extension package
- red before green: the clone-lifecycle tests were watched red against the unfixed source (four of six; the two edge assertions pin behavior that does not exist without the fix and pass vacuously without it). The strategy refactor itself is behavior-preserving, so its specs are green on both sides by design.
- e2e: left to CI

Deliberately untested: the hidden-tab path. Playwright's headed Chromium reports `visibilityState: "visible"` even with another page in front, so there is no honest end-to-end assertion for it; the listener wiring is covered by unit tests instead.

Found while probing and not addressed here: pausing the video triggers several full occlusion restarts, each re-initializing the mask provider. That is config churn upstream of this code and wants its own task.


Mr-Quin:
Task linked: [DA-679 Occlusion: retry cross-origin frame recovery instead of falsely reporting DRM](https://app.clickup.com/t/90131020449/DA-679)

PR #534, opened 2026-08-23
(extension) add agent verification tooling and spec theatre lint [DA-680]

Mr-Quin:
## Problem

Agent verification of extension work is self-reported. An agent that skips the browser produces output indistinguishable from one that did not, and a spec that asserts only on state passes while protecting nothing.

The build freshness contract also existed in only one place: `globalSetup` refused stale or wrong-env builds for the e2e suite, while nothing checked the artifact an agent explored by hand.

## Fix

- Extract the freshness rules into `e2e/setup/buildFreshness.ts`. `globalSetup` keeps throwing; the agent lane now enforces the same contract.
- Add `scripts/verify-preflight.ts` (`pnpm run verify:explore`): rebuilds an unambiguously stale build, hard-fails when chromium is missing, and writes a ready Playwright CLI config pointed at the same chromium the suite uses. It prints a fresh session name each run, because a reused browser profile can carry a wedged extension into the next run.
- Add `scripts/lint-specs.ts` (wired into `pnpm lint`): fails a spec that asserts only on state with no user-visible signal. Hard-fails locally where it bites the loop that wrote the spec; warns under `CI` so an outside contributor is never gated on it. Behavior with no UI surface opts out by naming the reason.
- Add tiered e2e scripts and tag the `@smoke` set.

`occlusion-cache-scope.spec.ts` already carried a written exemption; this makes that exemption machine-readable rather than inventing a second convention.

## Tests

- `pnpm lint` passes, including `lint:specs` across all 51 spec files.
- The lint was calibrated against the existing suite: its first version flagged three specs, all false positives (`toBeAttached`, a `waitForEvent('download'` with options, and the OPFS probe). Fixed the matchers, then confirmed it still fails a deliberately state-only spec and passes once that spec is removed.
- Freshness chain checked end to end: touching a source file makes `checkBuild` report stale, a direct `playwright test` then refuses to run, `verify:explore` rebuilds, and `e2e/specs/baseline/extension.spec.ts` passes (3 passed).
- `@smoke` tier: 3 passed in 6.7s.
- Full suite locally: 88 passed in 1.8 min, with one failure in `occlusion-cross-origin.spec.ts:154`. That test **passes in CI on this branch**, so it is a local-environment failure rather than a regression here or on master. Nothing in this PR touches occlusion.


Mr-Quin:
Task linked: [DA-680 Replace devtools MCP with Playwright CLI; make agent verification produce a spec](https://app.clickup.com/t/90131020449/DA-680)

PR #535, opened 2026-08-23
(chore) drop devtools MCP, make verification a spec that went red [DA-680]

Mr-Quin:
Stacked on #534.

## Problem

The devtools MCP was the only reason the agent lane touched `dev/chrome`, needed `CHROME_DEVTOOLS_EXECUTABLE`, and copied builds between worktrees. It also produced screenshots the agent chose, so verification evidence was still a self-report with pictures attached, and every selector had to be retyped from what the agent remembered seeing.

Three docs disagreed about which tests to run before pushing: the `e2e-spec` skill said full suite, `da-dev.md:73` said the suite should be running, and the standing agent memory said never run it locally.

## Fix

Replace the MCP with the Playwright CLI, which drives the same chromium the e2e suite uses. Parity was checked against a real build: extension load, service worker and all eight `__da` namespaces, popup render and drive, CDP `CSS.getPlatformFontsForNode`, console, network, screenshot. It adds video, tracing, and `generate-locator`, which turns a probed element into spec-ready code. Reloading the extension in place uses the same CDP `Extensions.loadUnpacked` call `swapExtension` already relies on.

- `browser-verify` is rewritten as exploration only, for when you don't yet know what to assert.
- `e2e-spec` and `e2e/AGENTS.md` make **red before green** the verification that counts: a spec has to fail against the build without the change. This replaces the mental mutate-the-assertion pass in `da-dev.md:121` with a run that actually has to fail. Refactors are named as the exception so nobody weakens an assertion to manufacture a red.
- Scratch specs live in `e2e/specs/.scratch/` (gitignored, skipped by the lint) and graduate by moving the file.
- Which-tests-to-run is settled with a measurement: the full suite is ~1.8 minutes, so it is the pre-push default and the tiers are for the inner loop. `--only-changed` selects on changed spec files, and specs load a built artifact rather than importing product source, so a content-script edit selects nothing.

## Tests

Docs and config only; no runtime code. The parity claims above were each executed against `packages/danmaku-anywhere/build` before writing them down, and the commands in the rewritten skill are the ones that were run.

### DIFF
--- a/.claude/commands/da-dev.md
+++ b/.claude/commands/da-dev.md
@@ -70,7 +70,7 @@ Make the changes. Apply these in order:
 | Packages      | `pnpm --filter <package> test`                     | N/A                                               |
 | Cross-cutting | `pnpm lint && pnpm --filter '...[origin/master]' test` | Depends on areas touched                          |
 
-Run e2e (`pnpm --filter @mr-quin/danmaku-anywhere test:e2e`) for any PR that adds or touches an e2e spec, or that changes content scripts, mount profiles, integration policies, dango manifests, or popup flows. The suite should be running before push.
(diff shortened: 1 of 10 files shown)


Mr-Quin:
Task linked: [DA-680 Replace devtools MCP with Playwright CLI; make agent verification produce a spec](https://app.clickup.com/t/90131020449/DA-680)

PR #536, opened 2026-08-23
(extension) run e2e headless by default and record verify-run video [DA-681]

Mr-Quin:
## Problem

**1. The suite ran headed for a reason that no longer holds.** `playwright.config.ts` carried `headless: false` with the comment *"Extensions require the full chromium channel and headed mode"*. The channel half is still true; the headed half was true of Chromium's old headless mode only.

**2. Verify runs never recorded video.** `video: isVerify ? 'on' : 'off'` was dead config. `use.video` only applies to contexts Playwright creates, and `e2e/setup/fixtures.ts` launches its own persistent context because extensions require one, so the setting was ignored. Video was named as the evidence artifact in #535 and produced nothing.

**3. No entry point for Playwright UI mode**, which is what you actually want when authoring a spec: watch mode, per-step DOM time travel, pick-locator, console and network panes.

**4. The headless default did not reach the migration lane.** `e2e/setup/swapExtension.ts` hardcoded `headless: false` in `launchExtension`, under a comment that only explains the CDP flag. So the swap spec still needed an XServer and "the suite no longer needs a display" was false as written. CI hid it by wrapping the run in `xvfb-run`; a dev machine hides it with a real display. On a bare container the spec dies with `Looks like you launched a headed browser without having a XServer running`.

**5. The suite silently depended on the ambient locale.** Chromium names a non-ASCII download `download` when the host locale is not UTF-8. Against the CJK season fixture that fails both `mount/export.spec.ts` specs with `Expected /\.xml$/i, Received "download"`. GitHub runners set `LANG=C.UTF-8`, so CI never saw it.

**6. `PLAYWRIGHT_HEADED: '1'` in the e2e workflow does nothing.** No Playwright version reads it (`PWDEBUG` is the real one; a grep across `playwright-core`, `playwright`, and `@playwright/test` finds zero references). It was already inert, but with headless as the default it reads as the thing keeping CI headed.

## Fix

- `headless: !process.env.DA_HEADED`. The suite no longer needs a display; `DA_HEADED=1` restores a window.
- Drop the `headless: false` override in `launchExtension`. Playwright injects config `use.*` into a direct `launchPersistentContext` call, so the config becomes the single source of truth and `DA_HEADED` now covers the migration specs too.
- `fixtures.ts` passes `recordVideo` at launch under `DA_VERIFY`, then attaches the finished `.webm` files, since a self-launched context is not attached to the report for us. The dead `use.video` is removed.
- `process.env.LANG ??= 'C.UTF-8'`, filled in only when absent, so download naming stops depending on the host.
- Drop `xvfb-run` and `PLAYWRIGHT_HEADED` from the e2e workflow.
- `test:e2e:ui` script, with the same `VITE_DA_ENV=e2e` prebuild hook as the other tiers.
- `--no-sandbox` from the verify preflight when running as root, so the explore lane starts inside a container.

## Tests

First commit:

- Full suite runs headless: confirmed across five runs, 86 to 89 passing depending on machine load. Part of that spread was problem 4, not load.
- Video: `DA_VERIFY=1` on `baseline/extension.spec.ts` produces 5 `.webm` and 3 traces, all attached to the HTML report. A normal run produces zero videos.
- `tsc --noEmit` clean, `pnpm lint` clean.
- UI mode was **not** exercised here. It needs a display and a controlling terminal and exits immediately when launched detached, so it is documented as human-only rather than claimed working.

Second commit, all on a container with no display:

- Full suite with no `DISPLAY` and no `LANG`: **89 passed in 2.5m**. The same command against the first commit: 3 failed (migration swap, both export specs).
- Migration specs headless: 3 passed in 30.5s, CDP `Extensions.loadUnpacked` included. With `DA_HEADED=1` and no display they fail on the XServer error, which is how we know the config now reaches that launcher.
- `DA_VERIFY=1`: 5 `.webm` recorded and attached. A normal run produces none and leaves no empty video dirs.
- `DA_HEADED=1` under xvfb: smoke band 3 passed.
- `tsc --noEmit` clean, `pnpm lint` clean, spec lint clean.

## What this does not claim

Headless is not a stability or speed fix. `occlusion-cross-origin` flakes locally under parallel workers in **either** mode; it passes run on its own (3 passed in 16.8s) and passes in CI. Reverting `fixtures.ts` to master's version and re-running headless reproduces the same failures plus one more, so the fixture change is not implicated. Local wall-clock varies more with machine load than with headless versus headed, so no speed claim is made.

Pinning `LANG` fixes the harness, not the underlying behavior. On Linux with a non-UTF-8 locale, a real export of a CJK-titled season still downloads as `download` with no extension. That is a product wart and out of scope here.


Mr-Quin:
Task linked: [DA-681 Run e2e headless by default, fix verify-run video, add UI mode script](https://app.clickup.com/t/90131020449/DA-681)

PR #537, opened 2026-08-24
(extension) assert on what rendered, not on a node that despawns [DA-682]

Mr-Quin:
## Problem

Three integration assertions raced a DOM node that is designed to disappear:

| Spec | Pattern |
| --- | --- |
| `unknown-mode-drop.spec.ts:91` | `playVideo()` then loop { seek; comment visible within **1s** }, 15s budget |
| `integration-auto-mount.spec.ts:83` | same |
| `integration-auto-mount.spec.ts:114` | same |

Rendered comments despawn as they scroll, so every check had to land inside a ~1s window after a seek. Under CI worker contention the render misses that window on every retry. `unknown-mode-drop` failed exactly this way on #536 and passed on a re-run of the same commit.

Worth noting what the suite does *not* have: there are no `waitForTimeout` sleeps anywhere. Waits are already `expect.poll` / `toPass`. The flakiness was never slow polling, it was that the only observable signal is transient.

**A second problem, quieter:** the malformed comment was asserted dropped via `toHaveCount(0)` on the DOM. A comment that rendered and scrolled away is indistinguishable from one never rendered, so that assertion could pass for the wrong reason and could not fail if the parser started rendering garbage.

## Fix

`DanmakuManager.service.ts` already runs a callback per rendered comment and receives the text. Record it on a content-script global (non-prod only, capped at 200), and read it back through the dev API exactly the way `__daMountMirror` is read: `chrome.scripting.executeScript` into the ISOLATED world. The player mounts in whichever frame holds the `<video>`, so this reads `allFrames: true`.

The assertions now ask what the renderer rendered. Each spec still asserts a visible signal, but on the **overlay** rather than an individual comment, because the overlay is what stays.

## Why a mirror is justified here

`e2e/AGENTS.md` puts the burden of proof on test-only mirrors and new dev-API methods. The case: the UI signal is transient *by design*, so it is genuinely unobservable at any later moment, and the mirror answers a question the DOM cannot answer at all — whether a comment was ever rendered. It makes the dropped-comment assertion strictly stronger rather than weaker. Precedent: 11 `IS_DA_E2E` branches already ship, including a mock mask provider in `uiIoc.ts`.

## Tests

- Negative control: with a deliberately wrong expectation the spec fails and prints the log, `["mode-2-comment","valid-comment"]` — the two valid comments present, the malformed one genuinely absent. The assertion bites, and the log is not empty.
- 54 repeats of the three specs after the fix: all passed.
- An earlier version that kept `commentElements().first()` as the visible signal failed **9 of 18** repeats, since asserting a comment node after the log confirmed the render is still a race. That is why the visible assertion is on the overlay.
- Runtime dropped from a 15s retry budget to ~3s per spec.
- `pnpm lint` clean, `tsc --noEmit` clean, smoke tier 3 passed.

## Rejected approach

Seeking while paused so `bindVideo` freezes the manager and comments stay on screen. Measured worse: 6/6 failures against 1/6 for the current code. `bindVideo.ts:230` explains why — `pushFlexibleDanmaku` aborts if `freeze` flips true before it commits, and the 200ms freeze timer beats the spawn under contention.

## Out of scope

`occlusion-cross-origin`, which flakes under load for a different reason (heavy MediaPipe/ONNX work under parallel workers).

### DIFF
--- a/packages/danmaku-anywhere/e2e/setup/da-client.ts
+++ b/packages/danmaku-anywhere/e2e/setup/da-client.ts
@@ -137,6 +137,20 @@ export class DaClient {
         timeoutMs,
       ] as const),
   }
+
+  danmakuRender = {
+    rendered: (tabId?: number): Promise<string[]> =>
+      this.sw.evaluate((id) => self.__da.danmakuRender.rendered(id), tabId),
+    waitForRendered: (
+      text: string,
+      tabId?: number,
+      timeoutMs?: number
+    ): Promise<string[]> =>
+      this.sw.evaluate(
+        ([x, id, t]) => self.__da.danmakuRender.waitForRendered(x, id, t),
+        [text, tabId, timeoutMs] as const
+      ),
+  }
 }
 
 export async function getDaClient(context: BrowserContext): Promise<DaClient> {
--- a/packages/danmaku-anywhere/e2e/specs/integration/integration-auto-mount.spec.ts
+++ b/packages/danmaku-anywhere/e2e/specs/integration/integration-auto-mount.spec.ts
@@ -21,8 +21,10 @@ const MOUNT_PATTERN = `${ORIGIN}/*`
 
 const EPISODE_TITLE = 'DA Integration Test'
 // Seeking past SEEK_TIME_S lands the cursor on at least one comment.
+// The comment due by SEEK_TIME_S; the render log is asserted against it.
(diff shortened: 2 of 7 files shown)


Mr-Quin:
Task linked: [DA-682 Replace transient-DOM races in e2e with a rendered-comment log](https://app.clickup.com/t/90131020449/DA-682)

PR #538, opened 2026-08-25
(proxy) cut per-request overhead on the hot path [DA-683]

Mr-Quin:
## Summary

The proxy is a pass-through to `ddp-microservice` but carries a fixed per-request tax. CPU p25 is 2982us against ddp's 531us, over ~105M requests/month, and its subrequest ratio is a flat 2.0 (service binding + a Sentry envelope on every request).

- Sample Sentry traces in production, `1.0` -> `0.001`. `captureException` is independent of trace sampling, so error reporting is unchanged.
- Resolve a Better Auth session only when the request carries a cookie or an `Authorization` header. A session cannot arrive any other way, so this cannot drop one.
- Gate hono's request logger to non-production, and drop the per-request logging in the cache middleware along with the now-unused `logPrefix` option.

No response shape or behavior change.

## Test plan

- lint (biome) pass; `pnpm --filter @danmaku-anywhere/proxy test` -> 13 files, 69 passed
- Deployed to staging and verified against `api.danmaku-staging.weeblify.app` with a real signed-in session: anonymous `/backup` 401; cookie 200; bearer 200; `get-session` returns the right user; `/ddp/api`, `/ddp/v1`, legacy `/proxy/api` 200; missing-path 400; 404 route 404.
- Measured the skip on staging worker analytics: a batch of anonymous requests ran at **1.00 subrequests/request**, a batch of credentialed ones at **2.00**. The extra subrequest is the Better Auth D1 lookup, so it is gone from the anonymous path.
- No e2e (backend-only, no extension surface).
- [ ] `pnpm --filter @danmaku-anywhere/proxy test`

## Notes

- Production loses hono's request lines. Workers Observability already head-samples production at 0.005, so only 0.5% were retained.
- An earlier commit moved Sentry scope enrichment into `onError` and was reverted: `@sentry/cloudflare` enables `honoIntegration` and `dedupeIntegration` by default, so the Hono integration captures before the app's own handler and dedupe discards the enriched event.
- `backend/proxy` is not type-checked in CI (DA-687). Checked by hand against a corrected config: no errors in the changed files.
- Follow-ups filed: DA-684, DA-685, DA-686.
- A test account exists on staging from the verification above (`da683-verify@example.com`).

### DIFF
--- a/backend/proxy/src/index.ts
+++ b/backend/proxy/src/index.ts
@@ -9,7 +9,7 @@ import { factory } from '@/factory'
 import { authContext } from '@/middleware/authContext'
 import { useCache } from '@/middleware/cache'
 import { requestLogger } from '@/middleware/requestLogger'
-import { applySentryContext, setContext } from '@/middleware/setContext'
+import { setContext } from '@/middleware/setContext'
 import { danDanPlay } from '@/routes/api/ddp/danDanPlay'
 import { llmLegacy } from '@/routes/api/llm/routes'
 import { getIsTestEnv } from '@/utils/getIsTestEnv'
@@ -86,7 +86,6 @@ app.onError((error, c) => {
 
   if (error instanceof HTTPException) {
     if (error.status >= 500) {
-      applySentryContext(c)
       Sentry.captureException(error)
     }
     return c.json(
@@ -100,7 +99,6 @@ app.onError((error, c) => {
     )
   }
 
-  applySentryContext(c)
   Sentry.captureException(error)
 
   const message =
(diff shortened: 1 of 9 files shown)


Mr-Quin:
Task linked: [DA-683 Cut proxy per-request overhead (Sentry sampling, hot-path logging, lazy auth)](https://app.clickup.com/t/90131020449/DA-683)

Mr-Quin:
Task linked: [DA-684 Front the ddp KV hot tier with the Cache API](https://app.clickup.com/t/90131020449/DA-684)
Task linked: [DA-685 Staging corpus gauge burns 77M D1 rows/day](https://app.clickup.com/t/90131020449/DA-685)
Task linked: [DA-686 Production D1 rows read crossed the included allowance](https://app.clickup.com/t/90131020449/DA-686)
Task linked: [DA-687 backend/proxy is not type-checked in CI](https://app.clickup.com/t/90131020449/DA-687)

PR #539, opened 2026-08-25
(chore) type-check backend/proxy in CI [DA-687]

Mr-Quin:
## Summary

CI runs `pnpm -r type-check`. `backend/proxy` had no `type-check` script, so it was silently skipped, and its `type-check:local` failed on the tsconfig before reading a line of source. Nothing type-checked the worker.

- Fix the tsconfig for the current tsgo: drop the removed `baseUrl`, make the path mapping relative, set `moduleResolution` explicitly.
- Point `types` at `@cloudflare/vitest-pool-workers/types`. That entry declares the `cloudflare:test` module; the bare package name resolves to the pool types, so every `cloudflare:test` import was unresolved.
- Rename the script to `type-check` so the recursive run picks it up.

The compiler then found two real defects:

- `middleware/rateLimit.ts` binds `DDP_RATE_LIMITER`, which exists in no environment (`ratelimits` is `[]` in all three). It has no callers and would have thrown had it gained one. Removed.
- The `ProvidedEnv` augmentation in `test-utils/mockBindings.ts` was inert, because `cloudflare:test` types `env` as `Cloudflare.Env`. The pool's `TEST_MIGRATIONS` binding is typed at its one consumer instead.

One change outside the package: `globalThis.crypto` is only typed under a DOM lib, so it broke the worker program through `integration-policy`. The bare `crypto` global is declared in the workers types and still throws a catchable `ReferenceError` where crypto is absent, so the surrounding try/catch behaves the same.

## Test plan

- `pnpm type-check` now reports `backend/proxy type-check: Done` instead of skipping it. 25 errors at the start, 0 now.
- Probe check: a middleware in `src/` reading `c.env.TEST_MIGRATIONS` fails to compile, confirming the test-only binding is not on the worker's `Env`.
- `pnpm --filter '...[origin/master]' test`: integration-policy 10 passed, danmaku-anywhere 771 passed, backend/proxy 69 passed.
- lint (biome) pass.
- [ ] `pnpm type-check`

## Notes

- No runtime behavior change in the worker beyond deleting an unreachable module.
- `include` stays `["./src", "./worker-configuration.d.ts"]`, so `vitest.config.ts` and `drizzle.config.ts` at the package root are still unchecked.
- `@danmaku-anywhere/integration-policy` resolves to source, so it now compiles under the worker's `lib: ["ESNext"]`. Future DOM-API use in a workspace package the worker imports will surface here rather than in that package.


Mr-Quin:
Task linked: [DA-687 backend/proxy is not type-checked in CI](https://app.clickup.com/t/90131020449/DA-687)

PR #540, opened 2026-08-26
(extension) resolve identityFields from the bundle so upgrades keep self-hosted namespaces [DA-692]

Mr-Quin:
Carries four related fixes for the upgrade path. The first is the release blocker; the other three were found while building the test for it.

## The blocker (DA-692)

A user upgrading from v1.5.0 with a self-hosted DanDanPlay config and downloaded seasons has those seasons silently rewritten onto the shared public `dandanplay` namespace, unrecoverably.

`ProviderService.setup()` fires the season identity reconcile on the same tick it registers `onInstalled`. A v1.5.0 profile has no manifest store, and `ManifestRegistry.init()` resolves `ready` off that empty store without seeding the bundled manifests, so `getIdentityFieldsMap()` returns `{}` and `computeNamespaceKey` falls through to returning `config.manifestId`. The reconciler stamps that and deletes `providerConfigId`, destroying the only handle that could correct the row.

The reconcile also races the provider-config option upgrade, so the first pass can see configs still in the v1.5.0 shape with no `manifestId` at all. Those rows are stamped `undefined` and orphaned permanently.

Offline this is deterministic. Online it is an uncontrolled race, so the same build corrupts one machine and not another. A mis-stamped row is byte-for-byte indistinguishable from a legitimately migrated one, so no later release can repair it.

`getIdentityFields` and `getIdentityFieldsMap` now fall back to the build-time bundled declaration for a manifest the registry has not loaded, and `identityFields` is optional through the reconciler: undefined means the manifest could not be resolved, and the row is skipped rather than stamped. `legacy:maccms` maps to an empty declaration explicitly, since it genuinely has none.

Chosen over eagerly seeding the store in `init()`, which would flip roughly 40 existing registry tests, make the seed-an-empty-store-as-preinstalled path dead code, and change the `kind` tag a fresh install shows before its first sync.

## Options migrations cannot wipe a store (DA-693)

`OptionsService.upgrade()` catches anything the chain throws and calls `reset()`. `extensionOptions` v23 assigned into `draft.playerOptions` unguarded, and `mountConfig` v2 through v4 called `data.map` directly. A corrupt entry inside the array did not throw at all: object spread tolerates `null`, so a bad record survived as a phantom config with no `patterns` or `name`.

Both now drop the bad record instead of resetting. `extensionOptions` and `danmakuOptions` had no test files at all; both chains now have coverage that runs the real upgrade closures through the real `migrateOptions` engine.

## Sources page crash (DA-694)

Opening the sources page before the registry finishes seeding threw `Cannot read properties of undefined (reading 'danmakuBaseUrl')`. `manifestById` comes from `listManifests()`, which awaits ready but not the bundle seed, so every config took the no-manifest branch of `versionSubtitle`, whose comment asserted only MacCMS could reach it.

## e2e cache slot (DA-695)

`ensureCurrentBuildForMigration` read `PLAYWRIGHT_WORKER_INDEX`, which Playwright does not set. The slot was always `0`, so parallel workers raced the same directory and failed with `EEXIST`. Now `TEST_PARALLEL_INDEX ?? TEST_WORKER_INDEX ?? '0'`.

## Test plan

- lint: pass. `pnpm type-check` pass, `biome check` clean on changed files, `lint-specs` passes.
- tests: 822 unit tests across 86 files, pass.
- e2e: `e2e/specs/migration/` 4 tests across 4 workers on a clean cache, pass. Full suite left to CI.
- red before green: observed, not reasoned.
  - Unit, against unmodified production code: `expected 'dandanplay' to match /^ns:/`, which is the defect verbatim, plus three others.
  - E2E: fix reverted and rebuilt, real v1.5.0 to HEAD CDP swap fails with `the self-hosted season kept its own namespace - Array [ "90001" ] + Array []`. Restored, rebuilt, passes.
  - The offline spec was proven red two independent ways: removing the bundled fallback, and commenting out `seedFromBundle()`.
  - Options guards: `TypeError: Cannot set properties of undefined` before, degrades to defaults after. The mountConfig corrupt-entry case failed as `expected length 2 but got 3`, the phantom config.
  - Crash guard: `TypeError` reproduced with the optional chaining removed.

## Notes for review

The migration fixture's self-hosted config now owns a season, seeded directly through IDB rather than the legacy popup, because v1.5.0's importer rewrites `providerConfigId` to the builtin id, so no UI route can express self-hosted ownership. That is a deliberate deviation from the seed-via-UI doctrine.

Residual risk not addressed here: if the catalog's `dandanplay` ever changed its `identityFields`, an early stamp from the bundle would disagree with the later-loaded manifest. `identityFields` is effectively a breaking-change field, and the mitigation is pinning the catalog to a tag rather than tracking a branch.

`legacy:maccms` mapping to an empty declaration preserves today's behavior but means multiple MacCMS instances continue to share one namespace. Giving MacCMS real identity fields would change existing MacCMS seasons' namespaces and orphan them, so that needs a migration rather than a flip. Filed separately.


Mr-Quin:
Task linked: [DA-692 Season identity is mis-stamped on upgrade when the manifest registry is empty](https://app.clickup.com/t/90131020449/DA-692)
Task linked: [DA-693 A throwing options migration step wipes the whole store](https://app.clickup.com/t/90131020449/DA-693)
Task linked: [DA-694 Sources page crashes during the manifest seeding window](https://app.clickup.com/t/90131020449/DA-694)
Task linked: [DA-695 e2e prior-release cache slot is shared across all workers](https://app.clickup.com/t/90131020449/DA-695)

PR #542, opened 2026-08-26
(extension) force-refresh the catalog past the browser and backend caches [DA-628]

Mr-Quin:
Supersedes #488, which could not be rebased: master's #524 changed `update()`'s return type to a three-state result while #488 changed the same method's signature, so the union had to be written by hand rather than resolved from conflict markers. Close #488 in favour of this.

Lets a user force a catalog refresh that actually reaches fresh bytes, past all three caches between the click and GitHub.

## There are three caches, not two

1. **The browser HTTP cache.** `Cache-Control: no-cache` as a request header only forces revalidation; it does not skip the cache. The forced path now also sets `cache: 'reload'` in the fetch init. `'reload'` over `'no-store'` because both skip the stored entry on the way out, but `'reload'` writes the fresh response back, so ordinary background syncs keep hitting the browser cache instead of becoming permanent full misses.
2. **The Worker's `caches.default`.** Already handled: `parseCacheControl` sets `noCache` for both `no-cache` and `max-age=0`, and the read is skipped when set. Both manifest routes share that middleware.
3. **Cloudflare's cache in front of the Worker's own outgoing `fetch()`.** Nothing bypassed this, so a forced refresh could skip the first two and still be served a copy up to five minutes stale, because the upstream sends `cache-control: max-age=300`. The user clicks refresh, everything reports success, and the bytes are identical.

Layer 3 is bypassed with `new Request(url, { cache: 'no-store' })`, which is the documented mechanism: the runtime will not check the cache and will not add the response to it, even when the response looks cacheable. Both routes go through one `fetchUpstream` helper sharing the same predicate as the middleware, so the two cannot disagree.

## Reconciling with #524

`update(entries?, force?)` keeps #488's parameters and master's `'unreachable' | 'empty' | 'synced'` return. Both failure paths still call `seedFromBundle()`, and `seedDefaultProviders()` stays outside the `'synced'` branch. #488's early-return on failure is gone, because that would have left a fresh offline user with manifests but no seeded provider configs, undoing #524.

`syncCatalog(force?)` loads the index once and passes it to all three steps, so a forced refresh issues one bypassing index fetch rather than three. `resolveCatalog()`'s undefined-versus-null distinction is preserved: undefined means fetch it, null means a lookup already failed, do not retry.

## Test plan

- lint: pass. `pnpm type-check` clean across all 11 projects.
- tests: backend 84 pass, extension unit 807 pass.
- e2e: catalog-refresh, updates, update-recovery, catalog-import, no-providers, 9 pass. Smoke 3 pass. Full suite left to CI.
- red before green: observed for each claim separately.
  - Backend subrequest bypass: `2 failed | 6 passed`, `expected 'https://raw.githubusercontent…' to be an instance of Request`.
  - Extension force plus coalescing, production reverted: `12 failed | 81 passed`, including `expected { cacheMode: undefined } to deeply equal { cacheMode: 'reload' }`.
  - **The #524 regression specifically**: moving `seedDefaultProviders()` back inside the `synced` branch and changing nothing else gives `3 failed`, the unreachable, empty and forced-unreachable seeding tests, all `expected "vi.fn()" to be called 1 times, but got 0 times`.
  - e2e against a build without the change: `2 failed`, `expect(forcedIndex).toHaveLength(1)` received `0`.

## Known limits and one risk

`cache: 'reload'` is invisible to Playwright route interception; the `Pragma: no-cache` header the Fetch spec describes arrives empty. So the browser-cache layer is provable only at the unit level, and `extensionFetchLike.test.ts` was added to pin that the `cache` mode survives the hop into global `fetch`. That seam was previously untested and is where this feature would silently die.

The e2e header no longer claims to prove a server-side bypass, since the route mock answers regardless of cache state. It names `cache.test.ts` and `manifest/router.test.ts` as where that is actually covered.

**Risk this introduces, not solved here:** `/manifest` and `/manifest/file` are mounted with no rate limiter, and the only limiter in the codebase is dead code with no binding in any environment. A forced refresh is triggerable by anyone sending `Cache-Control: no-cache` to a public endpoint, and it now costs one plus N unmetered requests to `raw.githubusercontent.com`. Tripping GitHub's limits would break the catalog for every user. Before this change that header did nothing to the subrequest, so this widens the surface. Mitigating: the response is still written into `caches.default`, so subsequent non-forced traffic stays cheap.

### DIFF
--- a/backend/proxy/src/middleware/cache.ts
+++ b/backend/proxy/src/middleware/cache.ts
@@ -81,6 +81,15 @@ function parseCacheControl(header?: string): RequestCacheControl {
   return directives
 }
 
+// Cloudflare caches a Worker's own outgoing fetch() subrequests, and that cache
+// is separate from caches.default: skipping the one below still leaves a route
+// free to serve bytes the origin's max-age is holding. Routes that proxy an
+// upstream use this to opt out of both for the same request.
+export function requestBypassesCache(req: HonoRequest): boolean {
+  const directives = parseCacheControl(req.header('Cache-Control'))
+  return Boolean(directives.noCache || directives.noStore)
+}
+
 export const useCache = (options: CacheOptions = {}) => {
   const { methods = ['GET'], maxAge, getCacheKey: customGetCacheKey } = options
 
--- a/backend/proxy/src/routes/api/manifest/router.test.ts
+++ b/backend/proxy/src/routes/api/manifest/router.test.ts
@@ -68,6 +68,49 @@ describe('Manifest API', () => {
     expect(content.id).toBe('builtin:bilibili')
   })
 
+  // The default-path assertions above pin a plain string URL, which is what
+  // keeps these two honest: an unconditional bypass would fail those.
+  it('bypasses the subrequest cache on a forced catalog fetch (GET /)', async () => {
+    const fetchSpy = vi.spyOn(globalThis, 'fetch').mockResolvedValueOnce(
+      new Response(JSON.stringify({ packageVersion: '1', manifests: [] }), {
+        status: 200,
+      })
+    )
+
+    const request = new Request(createTestUrl('/manifest'), {
+      headers: { 'Cache-Control': 'no-cache' },
+    })
+    await makeUnitTestRequest(request)
+
+    const upstream = fetchSpy.mock.calls[0][0] as Request
+    expect(upstream).toBeInstanceOf(Request)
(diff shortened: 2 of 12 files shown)


Mr-Quin:
Task linked: [DA-628 Force-refresh manifest catalog bypasses backend edge cache](https://app.clickup.com/t/90131020449/DA-628)

PR #543, opened 2026-08-26
(extension) gate the external RPC surface and narrow externally_connectable [DA-675]

Mr-Quin:
Closes the second route into the background RPC surface. PR #523 hardened the `window.message` bridge deliberately and did not seal this one.

## The hole

`externally_connectable.matches` was `*://danmaku.weeblify.app/*`. The scheme wildcard includes plaintext http, and that host sends no HSTS header on either the HTTPS response or the redirect, so a browser genuinely sends cleartext to it on a first visit, cleared state, or a new device.

`RpcManager` then bound the entire background RPC server to `chrome.runtime.onMessageExternal` with no filter. Roughly 78 commands, including `backupExport`, `cloudBackupList` and `cloudBackupDownload` riding the extension's own session, `dataWipeDanmaku`, `authDeleteAccount` with no re-auth, and `setHeaders`.

## The allowlist is two commands, and the evidence says it could be zero

`['mountConfigGetAll', 'mountConfigCreate']`.

The web app does not use this route at all. `grep -rn "chrome\." app/web/src` returns nothing; it detects the extension via a DOM attribute and talks over `window.postMessage`, which is the bridge #523 hardened. Kazumi search, chapter list and media extraction all live there.

The only in-repo external caller is `docs/src/components/onboarding/useConfig.tsx`, which sends exactly those two commands, and which is currently unreachable: `OnboardingAstro.astro` has no references, and the docs are served from `docs.danmaku.weeblify.app`, which the old pattern never matched. So the two admitted commands are the documented original purpose of the manifest key rather than anything with a live caller. Emptying the list is a one-line change if you want maximum lockdown.

Deleting `externally_connectable` outright is not the answer: omitting the key lets any other **extension** message the background, while specifying it with only `matches` blocks extensions. Narrowed is the correct shape.

## Mechanism

Gated by projecting the handler map, not by the RPC server's `filter` option. `filter`'s contract is "stay silent, this message is not mine", which the player frame routing depends on, so an external caller hitting a denied command would get an ambiguous port-closed failure instead of a clear error. `gateExternalHandlers` instead replaces every non-allowlisted method with a stub that throws a named error, and the ungated server no longer listens on `onMessageExternal`.

Default-deny by construction: a command added tomorrow is denied without anyone remembering.

## setHeaders

Denied externally, and separately scoped: `setRequestHeaderRule(rule, tabId)` now sets `tabIds: [tabId]` and the handler throws when `sender.tab?.id` is undefined, matching what `setMediaCorsRule` already did. Unscoped, that rule handed every tab in the browser a CORS bypass with a forced `Access-Control-Allow-Origin: *`.

No expiry, deliberately: DNR session rules have no TTL, and a `setTimeout` in an MV3 worker does not survive suspension, so an expiry would be a lie. `setMediaCorsRule` has the same lifecycle and relies on an explicit teardown. Giving `setHeaders` one means extending the app-extension contract, which is filed separately.

## Test plan

- lint: pass. `pnpm type-check` all 10 packages, `pnpm lint` clean, `lint-specs` ok.
- tests: extension 800 pass across 84 files, web-scraper 8 pass.
- e2e: `external-rpc.spec.ts` and `mount-config-crud.spec.ts` 3 pass, smoke 3 pass. Full suite left to CI.
- red before green: observed against a build of the merge base.

```
✘ external rpc: allowlisted command lands, the rest are rejected
    Expected: "errored"   Received: "success"     <- backupExport ran for an external page
✘ external rpc: a plaintext http origin cannot reach the extension
    Expected: false       Received: true          <- http://danmaku.weeblify.app reached the extension
```

Both pass after the change.

The http test navigates a real page to `http://danmaku.weeblify.app/` and calls `chrome.runtime.sendMessage`. Chrome only exposes `chrome.runtime` to origins matching an installed extension's `externally_connectable`, so this is the browser's own enforcement observed, not an assertion about config. Chromium only; Firefox has no `externally_connectable`, so there is nothing to verify there.

Regression guard is three-layered: unknown commands are denied by construction, the allowlist is pinned to an exact literal so widening it fails a test, and the e2e fails if anyone re-binds the ungated server.

## Notes

`appUrls.ts` turned out to take `isDev` as a parameter rather than reading `import.meta.env.DEV`, with a comment saying it must stay Node-importable, so the origin list is now de-duplicated through `getExternallyConnectablePatterns(isDev)` feeding both the manifest and the runtime constant.

Narrowing the patterns also gates `useExternalConnection`, a `window.message` handler in the controller content script accepting `hello` and `openPopup`, which is now off on plaintext too.

Not exercised end to end: the kazumi local-playback flow that consumes the header rule has no e2e coverage anywhere in the suite. The fetch happens in the same tab the content script sent from, so the scoping should hold, but the `messaging.ts` hunk is independently revertable if you want zero risk on this release.

### DIFF
--- a/packages/danmaku-anywhere/e2e/pom/ExternalCallerPage.ts
+++ b/packages/danmaku-anywhere/e2e/pom/ExternalCallerPage.ts
@@ -1,0 +1,61 @@
+import type { BrowserContext, Page } from '@playwright/test'
+
+const PAGE_BODY = `<!doctype html>
+<html lang="en">
+  <head><meta charset="UTF-8" /><title>DA external caller</title></head>
+  <body><p>External caller harness</p></body>
+</html>`
+
+export interface ExternalRpcOutcome {
+  // false when the browser refused the call outright, which is what a page on a
+  // non-connectable origin gets.
+  reached: boolean
+  state?: string
+  error?: string
+  output?: unknown
+}
+
+// A website driving the extension the way a real one does, through
+// chrome.runtime.sendMessage(extensionId, ...). The browser only exposes that
+// API to origins listed in the manifest's externally_connectable, so calls from
+// this page exercise the real boundary rather than a stubbed one.
+export class ExternalCallerPage {
+  private constructor(
+    private readonly page: Page,
+    private readonly extensionId: string
+  ) {}
+
+  static async open(
+    context: BrowserContext,
+    extensionId: string,
+    url: string
+  ): Promise<ExternalCallerPage> {
+    const page = await context.newPage()
+    await page.route(url, async (route) => {
+      await route.fulfill({
+        status: 200,
+        contentType: 'text/html; charset=utf-8',
+        body: PAGE_BODY,
+      })
+    })
+    await page.goto(url)
+    return new ExternalCallerPage(page, extensionId)
+  }
+
+  async call(method: string, input?: unknown): Promise<ExternalRpcOutcome> {
+    return this.page.evaluate(
+      async ({ extensionId, method, input }) => {
+        try {
+          const response = await chrome.runtime.sendMessage(extensionId, {
+            method,
+            input,
+          })
+          return { reached: true, ...response }
+        } catch (e) {
+          return { reached: false, error: e instanceof Error ? e.message : '' }
+        }
+      },
+      { extensionId: this.extensionId, method, input }
+    )
+  }
+}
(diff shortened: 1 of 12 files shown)


Mr-Quin:
Task linked: [DA-675 Narrow externally_connectable and gate the external RPC surface](https://app.clickup.com/t/90131020449/DA-675)

PR #544, opened 2026-08-26
(extension) loadable IoC containers and real test fakes [DA-689]

Mr-Quin:
## Problem

The Inversify containers cannot be loaded by a test.

- `background/ioc.ts` built a singleton `Container` at module scope.
- `common/ioc/uiIoc.ts` went further and performed I/O at import: it resolved `ExtensionOptionsService`, read `chrome.storage`, and called `i18n.changeLanguage` as a side effect of being imported.

So a test had two options: mock the container module away, or hand-construct the service with `new` and cast every collaborator through `unknown`. Both were taken. 16 of 71 injectable services had any test, 2 of 80 unit tests built a container, and no test anywhere validated the real wiring, so a missing binding or wrong scope was only discoverable by loading a browser.

The shared chrome mock made this worse: `onChanged.addListener` was a bare `vi.fn()`, so it could never emit. `ExtStorageService.setup()`, the cross-context sync between popup, worker and content scripts, was untestable, and no vitest config set `clearMocks` or `restoreMocks`.

## Fix

**Containers become modules.** Both split into a `ContainerModule` plus a `createXContainer()` factory, with the singleton exports preserved so the 15 consumer sites are unchanged. `uiIoc`'s import-time I/O moves into an explicit `bootstrapUiLanguage(container)`, called from the entrypoints that render translated UI. The standalone `StoreServiceSymbol` bindings move into the module itself so the exported module and the production container cannot diverge.

**A real fake replaces the mocks at the boundary.** `chrome` is stubbed with `@webext-core/fake-browser`: an in-memory implementation whose storage serializes as JSON like Chrome does, whose alarms are stateful, and whose events actually dispatch. `mockChromeApis.ts` loses its `@ts-nocheck` and is now a nine-line setup file.

Three deviations from the library are applied there, each for a production code path. `runtime.getBackgroundPage` is deleted, because an MV3 service worker has none and `isServiceWorker()` feature-detects on it. `permissions.contains` resolves `true`, because `validateOrigin` catches a throw and reports it as a malformed match pattern, which would otherwise fail every mount config URL. `webRequest` is left throwing, since nothing swallows that error and the failure should stay loud.

**Shared test layer.** `createTestContainer` loads production modules with per-token overrides via `rebindSync`; one `silentLogger` replaces eight hand-rolled copies across eleven files; typed factories replace the local `makeSeason` duplicates in the persistence tests.

**Isolation is now real.** `clearMocks`, `restoreMocks`, `unstubEnvs` and `unstubGlobals` are on, and the fake is reset before every test. Because the fake's methods are plain functions rather than module-scope `vi.fn(impl)` wrappers, `vi.spyOn` plus `restoreMocks` restores them without any explicit per-method bookkeeping.

No production behavior changes beyond the two narrow ones above (explicit language bootstrap, standalone bindings location).

## Tests

- 812 extension unit tests pass, up from 794 on master. Full monorepo suite, `type-check`, `lint` and `lint:specs` all clean; both production builds succeed. Lint warning count is unchanged from master's baseline of 16.
- New wiring tests resolve every service the background entrypoint resolves, assert all nine `StoreServiceSymbol` services in order, assert singleton identity, and assert both factories are callable. They fail on a missing binding or a throwing constructor.
- `uiIoc.test.ts` additionally guards that importing the module performs no `chrome.storage` access.
- Specs that previously stubbed `chrome.storage` now seed it and assert the resulting stored state, and the `onInstalled` seeding tests dispatch the real event instead of retrieving the listener from a mock's call list.
- Regression tests cover bootstrap idempotence, container-module and production divergence, and overriding an autobound class token from the real background module.
- `content/app/index.test.ts` no longer mocks a container module, so the app-bridge origin checks now run against real wiring.

Each migrated assertion was checked by breaking the code it covers and confirming the test fails: the ManifestStore write mutex and its absent-key guard, `ExtStorageService.delete()`, the `onInstalled` registration, the v15 upgrade reading storage, the import-time language bootstrap, and refresh-alarm creation.

Zero skipped tests before and after.

Follow-ups tracked in DA-690 (typed doubles, theatre removal, lint enforcement) and DA-691 (service-layer backfill, CI holes, coverage).


Mr-Quin:
Task linked: [DA-689 Phase 1: loadable containers + test infrastructure](https://app.clickup.com/t/90131020449/DA-689)
Task linked: [DA-690 Phase 2: typed doubles and theatre removal](https://app.clickup.com/t/90131020449/DA-690)

PR #545, opened 2026-08-30
(extension) typed doubles for unit test collaborators [DA-690]

Mr-Quin:
## Summary
- Replace `as unknown as X` / `as any` casts on test-double collaborators with `satisfies Pick<X, ...>` or `vi.fn<X['method']>()` so a rename or signature change fails at compile time instead of doing nothing silently.
- Rewrite `ProviderService.test.ts` (56 casts, 8 positional constructor args), `AlarmManager.test.ts`, and `ProviderFactory.test.ts` to build the service under test through `createTestContainer` with per-token overrides instead of positional `new` construction or a hand-rolled DI context.
- Add a unit-test pass to `scripts/lint-specs.ts`, run as part of `pnpm lint`, failing on: `as unknown as` without a documented `lint-specs-allow-cast:` reason, a test whose assertions are exclusively `toHaveBeenCalled*` without `lint-specs-allow-call-only:`, and a `vi.mock` of a module neither the test file nor the code under test imports.
- Delete `BackupService.test.ts`'s three delegation-only tests (they only asserted a one-line pass-through, behind an unused `vi.mock`); replace with real coverage of `importAll`'s JSON-parse branch, the one piece of actual logic in that file. `ConfigStateService.test.ts` already covers `restoreState`'s real logic and loses its own dead `vi.mock` calls.
- Rewrite `ExtStorageService.test.ts` to exercise the cross-context storage sync path (`setup()`/`destroy()` against a real `storage.onChanged` listener, `read()` on a missing key) instead of round-tripping the same value back through `chrome.storage`.
- Remaining casts that can't be safely typed (generic methods like `SeasonService.findExisting`/`DanmakuService.upsert`, external DOM/browser API fixtures like `HTMLVideoElement`/`chrome.runtime.Port`, deliberately invalid inputs for fallback-path tests) each carry an inline `lint-specs-allow-cast:` comment naming the reason.

## Test plan
- Verified: lint pass (`tsc` + `biome` + `lint-specs`) · `pnpm --filter @mr-quin/danmaku-anywhere test` → 816 passed (87 files) · e2e not run locally (test-layer refactor only, no runtime behavior change; e2e is CI's job)
- Reintroduced a bare `as unknown as` cast and confirmed `lint-specs.ts` fails on it (exit 1), then reverted.
- Broke `ExtStorageService.setup()`'s listener registration and confirmed the new cross-context sync test fails, then reverted.
- Broke `BackupService.importAll`'s JSON-parse throw and confirmed the new test fails, then reverted.
- Widened `lint-specs.ts`'s call-only-assertion check to also recognize `expect(...).rejects.`/`.resolves.` chains, which it previously missed entirely.

## Notes
- No production code changed; only test files and the lint-specs dev-tooling script.
- Coverage threshold gating is out of scope for this PR (tracked separately in the epic).


Mr-Quin:
Task linked: [DA-690 Phase 2: typed doubles and theatre removal](https://app.clickup.com/t/90131020449/DA-690)

PR #546, opened 2026-08-31
(extension) service-layer test backfill, CI test-script gate, coverage report [DA-691]

Mr-Quin:
## Summary
- Backfills tests for the five highest-risk untested services: the extension-options, mount-config and danmaku-options migration chains (field-level assertions per version step plus the throw-resets-store path), `DanmakuService.import` against a real Dexie, an `RpcManager` command-surface harness, and the three episode-matching strategies.
- The `RpcManager` harness pins the 85 externally-reachable commands and the 21 internal-only relay commands as separate sets, and asserts which listen target each server gets. A command added to the wrong map, or moved between them, fails the test.
- Fixes a data-loss bug the backfill surfaced. The extension-options version 21 step migrates `danmakuSources` into the provider store, but that store does not exist yet when the step runs. `OptionsService.set()` reads the missing store for its version and rejects, and the call site discarded the promise inside a `try/catch` that could never fire, so every migrated provider config was silently dropped. The write now names its version explicitly.
- `pnpm -r test` skips a package with no `test` script without saying so. `app/web` (script was named `test:ng`), `bangumi-api` and `result` all now declare one, and `pnpm check:test-scripts` fails if a future package does not.
- Adds `@vitest/coverage-v8` and `test:coverage`, printed in CI. No threshold gate, per the plan to add a ratchet once the service layer is covered. Baseline: 28.78% statements, 19.65% functions.
- Resolves the two doc contradictions: the web app is Vitest via the `@angular/build:unit-test` builder, not Jasmine+Karma, and the test-header rule now reads the same in `CLAUDE.md` and `packages/danmaku-anywhere/AGENTS.md` (required for e2e, optional for unit and integration).

## Test plan
- Verified: lint pass (tsc + biome + lint-specs) · tests `pnpm --filter '...[origin/master]' test` -> all green, extension 882 passed across 95 files, app/web 21, backend 82, packages 156 · e2e skipped, no runtime behavior on a page changed
- Red before green, each confirmed by reverting the change or mutating the code under test: the version 21 fix (2 tests), gutting versions 22 through 27, replacing `migrateDanmakuSourcesToProviders` with its defaults fallback, dropping the MacCMS record in provider migration step 2, removing the version 25 hotkey merge, swapping the title-rule and mapping-rule order in `LocalMatchingStrategy`, forcing its `matchLocalDanmaku` guard, and replacing `matchLocalByTitle`'s path matching with "return the first episode". Removing a package's `test` script makes `check:test-scripts` exit 1.
- [ ] `pnpm check:test-scripts` from the repo root and from inside a package directory
- [ ] `pnpm --filter @mr-quin/danmaku-anywhere test:coverage`
- [ ] `cd app/web && pnpm test` should exit rather than enter watch mode

## Notes
- Ordering matters here and is easy to get backwards. `OptionsService.set()` waits on readiness, and `UpgradeService` opens readiness only after every store has upgraded, so the version 21 write lands *after* the provider store has already upgraded and reset itself to defaults. It has to win that race, and the version it names is the one the store is left on. The handoff test leaves readiness closed so it reproduces that path rather than the easier pre-ready one, and a second test runs the upgrade twice to show the configs survive.
- That version is a literal rather than a "latest" constant, so a future provider migration still picks this data up instead of stepping over it.
- `ng test` defaults `--watch` to TTY presence, so the renamed `app/web` `test` script would have hung in an interactive shell. Pinned `--watch=false`.
- The CI test step now excludes the extension and a following step runs it with coverage, so the extension suite runs once rather than twice while the report still prints on every PR. Confirmed the negative filter excludes from the positive selection rather than unioning with it.
- Version 21 only runs for installs still below it, and the migration drops the `iqiyi` source because it is not a built-in provider. That is existing behavior, now asserted rather than assumed.


Mr-Quin:

